PR #16, opened 2020-06-02
Issue-9-Moving the storage account creation from constructor to own method.

AnoojNair:
Hi @aslotte ,

Had to remove the readonly as it can only be set in a constructor and not in a method.

### DIFF
--- a/MLOps.NET/MLLifeCycleManager.cs
+++ b/MLOps.NET/MLLifeCycleManager.cs
@@ -7,11 +7,18 @@ namespace MLOps.NET
 {
     public class MLLifeCycleManager
     {
-        private readonly IMetaDataStore metaDataStore;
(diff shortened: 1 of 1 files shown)


MLOps.NET/MLLifeCycleManager.cs
```diff
@@ -7,11 +7,18 @@ namespace MLOps.NET
 {
     public class MLLifeCycleManager
     {
-        private readonly IMetaDataStore metaDataStore;
-
-        public MLLifeCycleManager(string connectionString)
+        // No readonly as a readonly field can only be set in a constructor.
```
aslotte, line 10: Please remove this comment

aslotte:
@AnoojNair thank you for your contribution! We'll go ahead and merge this and continue working on separating this out into it's own package.

PR #20, opened 2020-06-02
Refactor entities to interfaces

dcostea:
We need to extract the interfaces from entities in order to prepare the code for adding another storage (SQLite).

### DIFF
--- a/MLOps.NET.Azure/Entities/Experiment.cs
+++ b/MLOps.NET.Azure/Entities/Experiment.cs
@@ -17,7 +17,5 @@ namespace MLOps.NET.Azure.Entities
         public Guid Id { get; set; }
 
         public string ExperimentName { get; set; }
-
-        public IList<IRun> Runs { get; set; }
     }
 }
--- a/MLOps.NET.Azure/Entities/Experiment.cs
+++ b/MLOps.NET.Azure/Entities/Experiment.cs
@@ -1,10 +1,11 @@
 ﻿using Microsoft.Azure.Cosmos.Table;
+using MLOps.NET.Entities.Entities;
 using System;
 using System.Collections.Generic;
 
 namespace MLOps.NET.Azure.Entities
 {
-    internal sealed class Experiment : TableEntity
+    internal sealed class Experiment : TableEntity, IExperiment
     {
         public Experiment(string experimentName)
         {
@@ -17,6 +18,6 @@ namespace MLOps.NET.Azure.Entities
 
         public string ExperimentName { get; set; }
 
-        public List<Run> Runs { get; set; }
+        public IList<IRun> Runs { get; set; }
     }
 }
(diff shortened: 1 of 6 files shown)


aslotte:
Thanks @dcostea! 

I added a comment I would like to get your thoughts on before merging this.

dcostea:
I'm afraid I don't know where to look for comments (to discuss them). It's the first time I'm using GitHub interface. Or they are not there yet?

MLOps.NET.Azure/Entities/Experiment.cs
```diff
@@ -17,6 +18,6 @@ public Experiment(string experimentName)
 
         public string ExperimentName { get; set; }
 
-        public List<Run> Runs { get; set; }
+        public IList<IRun> Runs { get; set; }
```
aslotte, line 21: Sorry about that, the comment appears to have disappeared. The comment was about this property. It should probably actually be removed as we don't use it in the NoSQL scenario. That leaves the question though how we do with other implementations that need it. I see it that we can either:

1. Keep our entities flat
2. Keep this property but populate it in the NoSQL case after fetching an experiment

dcostea: We might segregate the interface, therefore we implement only those that are needed?

Just to double-check (because I'm missing the big picture here), is that making sense if I split IExperiment into IExperiment (without Run) and IRunableExperiment (having Run)?

aslotte: I think they all in theory have runs, just not as part of their composition :)
To keep it simple now, could we make them flat? E.g. not having a list of Runs?

dcostea: but only one? like public IRun Run { get; set;} ?

aslotte: All runs would have an associated ExperimentId, so we can fetch them using that instead. I think we probably also want to add ExperimentId as a field to IRun, and RunId to IMetric

aslotte: Happy to open a PR to this PR with the changes I'm suggesting if that helps to clarify it

dcostea: Oh, first take a look at my latest commit.

aslotte: Love it, this is exactly what I was thinking we could start with!

aslotte: One last thing @dcostea and then I think we're good, could you set the ExperimentId in the Run class' constructor?

dcostea: done!

dcostea:
> LGTM! 👍 Thank you for the contribution @dcostea and welcome to the group of contributors!

Thank you!

PR #30, opened 2020-06-03
Add sqlite

dcostea:
I had to add an Id property for the Metric entity, as well. I need it for SQLite db. I hope is not a problem.


MLOps.NET.SQLite/Entities/Experiment.cs
```diff
@@ -0,0 +1,18 @@
+﻿using MLOps.NET.Entities.Entities;
+using System;
+
+namespace MLOps.NET.SQLite.Entities
+{
+    public sealed class Experiment : IExperiment
```
aslotte, line 6: We can probably make this internal

MLOps.NET.SQLite/Entities/Metric.cs
```diff
@@ -0,0 +1,21 @@
+﻿using MLOps.NET.Entities.Entities;
+using System;
+
+namespace MLOps.NET.SQLite.Entities
+{
+    public sealed class Metric : IMetric
```
aslotte, line 6: Same here in regards to internal

MLOps.NET.SQLite/Entities/Run.cs
```diff
@@ -0,0 +1,23 @@
+﻿using MLOps.NET.Entities.Entities;
+using System;
+
+namespace MLOps.NET.SQLite.Entities
+{
+    public sealed class Run : IRun
```
aslotte, line 6: Same here in regards to internal

MLOps.NET.SQLite/Storage/LocalDbContext.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.EntityFrameworkCore;
+using MLOps.NET.SQLite.Entities;
+
+namespace MLOps.NET.SQLite.Storage
+{
+    public class LocalDbContext : DbContext
```
aslotte, line 6: I'm not a SQLite expert, but would it make sense not to limit this to Local databases? E.g. I would image an end user probably wanting to have a central place for storage that may not be the same resource as the model is trained on.

MLOps.NET.SQLite/Storage/SQLiteMetaDataStore.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using System.Threading.Tasks;
+
+namespace MLOps.NET.Storage
+{
+    public sealed class SQLiteMetaDataStore : IMetaDataStore
+    {
+        public async Task<Guid> CreateExperimentAsync(string name)
+        {
+
```
aslotte, line 12: Extra newline

MLOps.NET.SQLite/Storage/SQLiteMetaDataStore.cs
```diff
@@ -0,0 +1,47 @@
+﻿using MLOps.NET.SQLite.Entities;
+using MLOps.NET.SQLite.Storage;
+using System;
+using System.Threading.Tasks;
+
+namespace MLOps.NET.Storage
+{
+    public sealed class SQLiteMetaDataStore : IMetaDataStore
```
aslotte, line 8: Could this be made internal?

MLOps.NET.SQLite/Storage/LocalDbContext.cs
dcostea: SQLite is file-based whereas MSSQL Server and MySQL are server-based. I think it is not meant for concurrence, therefore it is better to be local.
For the client-server scenario, we can use MSSQL, but in storage project.

aslotte: Gotcha, that was the missing piece of the puzzle, thanks!

aslotte:
@dcostea I couldn't see an extension method to instantiate the SQLiteMetaDataStore. You may want to add one, e.g. `UseSQLite()`.

Would you be able to test it real quick locally as well to make sure it appears to work? If not, I'm happy to try it during one of the streams. Thanks!

dcostea:
> @dcostea I couldn't see an extension method to instantiate the SQLiteMetaDataStore. You may want to add one, e.g. `UseSQLite()`.
> 
> Would you be able to test it real quick locally as well to make sure it appears to work? If not, I'm happy to try it during one of the streams. Thanks!

I have added UseSQLite. I will try to test later, otherwise, we can do it in the next stream.

PR #31, opened 2020-06-04
added function to upload the model artifact to azure blob storage

### LINKED ISSUE #3: Add support to log an artifact during run
During a run we want to upload a model to a container in Azure.

I'm envisioning the structure to look like a container name ModelRepository and each model being named the unique GUID for the run.

AnoojNair:
Fixes #3 

Contains the changes to upload the model artifact to a container in azure blob storage.

1. Container name is 'model-repository'. If it doesn't the method creates it first.
2. The blob name is the runId with a .zip extenion.
3. Doesn't apply any permissions on the blob.

### DIFF
--- a/MLOps.NET.Azure/Storage/StorageAccountModelRepository.cs
+++ b/MLOps.NET.Azure/Storage/StorageAccountModelRepository.cs
@@ -9,18 +9,18 @@ namespace MLOps.NET.Storage
     internal sealed class StorageAccountModelRepository : IModelRepository
     {
         private readonly BlobContainerClient blobContainerClient;
-        private const string _containerName = "model-repository";
-        private const string _fileExtension = ".zip";
+        private const string containerName = "model-repository";
+        private const string fileExtension = ".zip";
 
         public StorageAccountModelRepository(string connectionString)
         {
-            this.blobContainerClient = new BlobContainerClient(connectionString, _containerName);
+            this.blobContainerClient = new BlobContainerClient(connectionString, containerName);
         }
 
         public async Task UploadModelAsync(Guid runId, string filePath)
         {
             await this.blobContainerClient.CreateIfNotExistsAsync();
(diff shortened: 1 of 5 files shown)


MLOps.NET.Azure/Storage/StorageAccountModelRepository.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace MLOps.NET.Storage
+{
+    internal sealed class StorageAccountModelRepository : IModelRepository
+    {
+        private readonly BlobContainerClient blobContainerClient;
+        private const string _containerName = "model-repository";
```
aslotte, line 12: My bad, I should have added a coding style guide to the repo. Just to keep it consistent I think we will not be using underscore for private member variables (see blobContainerClient above).

MLOps.NET/MLLifeCycleManager.cs
```diff
@@ -29,9 +31,21 @@ public async Task LogMetricAsync(Guid runId, string metricName, double metricVal
             await MetaDataStore.LogMetricAsync(runId, metricName, metricValue);
         }
 
+        /// <summary>
```
aslotte, line 34: I'm not against summary comments, but we should probably be consistent (meaning I should probably update the other methods too... :) 

With that said, this comment should probably be made more generic as what we are doing is just storing the model in a repository. In the future that can be other places than Azure, e.g. on a local fileshare

AnoojNair: Ohh yes,  have updated the comment to be more generic 👍

MLOps.NET.Azure/Storage/StorageAccountModelRepository.cs
AnoojNair: Done 👍

Issue #35, opened 2020-06-04
Add NuGet package information to MLOps.NET.SQLite


aslotte:
We need to add
- License
- Version number

Anything else required to be in line with other packages.

ssa3512:
This should be resolved now.

PR #50, opened 2020-06-06
Add SourceLink and GitVersioning

### LINKED ISSUE #42: Add CI/CD pipeline for NuGet packaging and deployment
### Background
We need to automate the build and deployment of our nuget packages.
Each package needs to have a consistent build number.

We should probably run the dotnet pack command with versioning after each CI build (we can have different versions for PR builds vs master builds).

The deployment to nuget.org should most likely be a separate workflow file as it should be manually triggered once we have a release.

ssa3512:
## Resolves
Resolves #42, #35 

## Description
This PR adds a Directory.Build.props, Directory.Build.targets and version.json in the root of /src to add sourcelink, gitversioning and deterministic builds to MLOps.NET

Documentation on GitVersioning: https://github.com/dotnet/Nerdbank.GitVersioning
Documentation on Deterministic builds is here: https://github.com/clairernovotny/DeterministicBuilds

## TODO:
- [x] Publish PR packages - might need @aslotte to set this up (requires a package feed and keys) edit: publishing as build/CI artifact
- [x] Fix deterministic portion (nuget package explorer shows untracked sources warning and non-deterministic

### DIFF
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -13,6 +13,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -34,6 +34,6 @@ jobs:
         name: nuget-packages
         path: '*.nupkg'
     - name: Push Packages
-      if: ${{ github.ref }} == 'master' && ${{ github.event_name }} != 'pull_request'
+      if: ${{ github.ref == 'master' && github.event_name != 'pull_request }}'
       run: dotnet nuget push **/*.nupkg --skip-duplicate --no-symbols true
     
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -19,9 +19,6 @@ jobs:
       uses: actions/setup-dotnet@v1
       with:
         dotnet-version: 3.1.101
-        source-url: https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json
-      env: 
-        NUGET_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
     - name: Install dependencies
       run: dotnet restore src/MLOps.NET.sln
     - name: Build
@@ -33,7 +30,3 @@ jobs:
       with:
         name: nuget-packages
         path: '*.nupkg'
-    - name: Push Packages
-      if: ${{ github.ref }} == 'master'
-      run: dotnet nuget push *.nupkg --skip-duplicate --no-symbols true
-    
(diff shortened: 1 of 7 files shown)


ssa3512:
@aslotte I enabled `<GenerateDocumentationFile>` with this so the nuget package contain an xml doc file (for intellisense etc) - this generates a bunch of CS1591 warnings for all public members that do not have doc comments. Do you prefer these be suppressed or would you want to keep them so they can be addressed in a separate issue?

aslotte:
That's awesome @ssa3512, thank you for looking into this, let me open a new issue for comments on public properties/methods.

aslotte:
Created #52 to track this. Let me know if you have anything else to add there.

aslotte:
Added interface summary comments to #54 that was just closed

aslotte:
What can I do next before we can merge this? In your experience, does it make sense to publish PR nuget packages to something like GitHub nuget packages, or just leave them as artifacts?

ssa3512:
> What can I do next before we can merge this? In your experience, does it make sense to publish PR nuget packages to something like GitHub nuget packages, or just leave them as artifacts?

I don't know that it really makes sense to publish PR packages to a feed as they aren't really meant for consumption. From what I have seen it makes sense to have CI packages go to a separate feed (myget/azure artifacts/github/etc) and then you can publish beta, rc, and release to nuget.org.

I would recommend a second job using [jobs.<job.id>.if](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) for publishing CI pacakges and probably a manually triggered job that is locked down to only you for pushing to nuget.org

ssa3512:
I am having trouble getting packages to push to github package registry correctly, even in my own test repository. I will have to dig into it some further. It looks like there are some caveats around using GPR nuget and github actions.

aslotte:
Thanks for looking into this @ssa3512. I haven't worked much with CI/CD on nuget packages so I appreciate your input. I'll try to get up to speed in the next few days.

ssa3512:
This should be ready now - will always create packages and publish as an artifact, but will only push to your repository's nuget feed for CI on master.

You can see an example of the published packages here: https://github.com/ssa3512/MLOps.NET/packages/258569

If you open with Nuget Package Explorer they light up green for sourcelink and deterministic

image

To change the version, you just update src/version.json. The third digit in the package is always git-height since the last version change.

aslotte:
@ssa3512 this is great, thank you for the explanation as well. I've learned a lot personally by reviewing this contribution. Let me get this merged and set up my access token in secrets.

ssa3512:
@aslotte I don't think you need to set up an access token - because this runs in the repository GITHUB_TOKEN is automatically available. https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#about-the-github_token-secret

aslotte:
I just saw, it's magic :)

aslotte:
Question @ssa3512 - we would expect a package to be pushed on merge to master during the CI build right? Just wondering as I don't see any packages so far. The artifacts seem to be created, but they are missing the git suffix

Issue #51, opened 2020-06-06
SQLite is not making use of Data folder to store the db file


dcostea:
**Describe the bug**
SQLite is not making use of Data folder to store the db file.

**To Reproduce**

**Expected behavior**
Create the db in the Data folder

aslotte:
@dcostea can this be closed?

dcostea:
yes! thank you.

PR #55, opened 2020-06-06
Add file storage for SQLite. Add integration tests.

### LINKED ISSUE #34: Test SQLite implementation
@dcostea adding this issue so we remember to test it at some point :)

dcostea:
## Resolves
#34 #51 

## Description

I have added two integration tests (one for creating an experiment in sqlite db and another one for uploading a model file)

### DIFF
--- a/src/MLOps.NET.SQLite.IntegrationTests/SQLiteMetaDataStore.cs
+++ b/src/MLOps.NET.SQLite.IntegrationTests/SQLiteMetaDataStore.cs
@@ -1,0 +1,46 @@
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.IO;
+
+namespace MLOps.NET.SQLite.IntegrationTests
+{
+    [TestClass]
+    public class SQLiteMetaDataStore
+    {
+        [TestMethod]
+        public void CreateExperimentAsync_Success()
+        {
+            //Arrange
+            MLLifeCycleManager mlm = new MLLifeCycleManager();
+            mlm.UseSQLite();
+
+            //Act
+            var guid = mlm.CreateExperimentAsync("first experiment").Result;
+
+            //Assert
+            Guid.TryParse(guid.ToString(), out var parsedGuid);
+            parsedGuid.Should().NotBeEmpty();
+        }
+
+        [TestMethod]
+        public void UploadModelAsync_Success()
+        {
+            //Arrange
+            MLLifeCycleManager mlm = new MLLifeCycleManager();
+            mlm.UseSQLite();
+            var guid = Guid.NewGuid();
+            var modelPath = @"C:\data\model.zip";
+            var modelStoragePath = @"C:\MLOps";
+            using var writer = new StreamWriter(modelPath);
+            writer.Close();
+
+            //Act
+            mlm.UploadModelAsync(guid, modelPath).Wait();
+
+            //Assert
+            var fileExists = File.Exists(Path.Combine(modelStoragePath, $"{guid}.zip"));
+            fileExists.Should().BeTrue();
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/MLOps.NET.SQLite/Storage/LocalFileModelRepository.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.IO;
+using System.Threading.Tasks;
+
+namespace MLOps.NET.Storage
+{
+    internal sealed class LocalFileModelRepository : IModelRepository
+    {
+        public async Task UploadModelAsync(Guid runId, string filePath)
+        {
+            var modelStoragePath = @"C:\MLOps";
```
aslotte, line 11: @dcostea could we make this a variable passed into UseSQLite? I envision that one may want to store this on e.g. a fileshare or something else, so it would be awesome if it was configurable (we can let it default to C:\MLOps though so you don't have to set it)

src/MLOps.NET.SQLite.IntegrationTests/SQLiteMetaDataStore.cs
```diff
@@ -0,0 +1,46 @@
...
+            MLLifeCycleManager mlm = new MLLifeCycleManager();
+            mlm.UseSQLite();
+            var guid = Guid.NewGuid();
+            var modelPath = @"C:\data\model.zip";
+            var modelStoragePath = @"C:\MLOps";
+            using var writer = new StreamWriter(modelPath);
+            writer.Close();
+
+            //Act
+            mlm.UploadModelAsync(guid, modelPath).Wait();
```
aslotte, line 39: Let's make the test method async and use await here

src/MLOps.NET.SQLite.IntegrationTests/SQLiteMetaDataStore.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace MLOps.NET.SQLite.IntegrationTests
+{
+    [TestClass]
+    public class SQLiteMetaDataStore
+    {
+        [TestMethod]
+        public void CreateExperimentAsync_Success()
+        {
+            //Arrange
+            MLLifeCycleManager mlm = new MLLifeCycleManager();
```
aslotte, line 15: Super awesome that you're adding the first tests! For a convention, I think it would be nice if we call the class we are testing either classUnderTest or unitUnderTest, I'll let you decide between those two :)

src/MLOps.NET.SQLite/Storage/LocalFileModelRepository.cs
dcostea: it makes sense

src/MLOps.NET.SQLite.IntegrationTests/SQLiteMetaDataStore.cs
dcostea: I'm not sure this is what you meant, I put CreateExperimentAsync_Always_ReturnsNonEmptyGuid and UploadModelAsync_ValidModelPath_UploadSuccess

src/MLOps.NET.SQLite/MLLifeCycleManagerExtensions.cs
```diff
@@ -4,10 +4,10 @@ namespace MLOps.NET.SQLite
 {
     public static class MLLifeCycleManagerExtensions
     {
-        public static MLLifeCycleManager UseSQLite(this MLLifeCycleManager mLLifeCycleManager)
+        public static MLLifeCycleManager UseSQLite(this MLLifeCycleManager mLLifeCycleManager, string destrinationFolder)
```
aslotte, line 7:
```suggestion
        public static MLLifeCycleManager UseSQLite(this MLLifeCycleManager mLLifeCycleManager, string destrinationFolder = "C:\MLOps")
```

PR #56, opened 2020-06-06
Fix duplicate experiment creation

### LINKED ISSUE #48: CreateExperiment creates duplicates in Azure Storage
**Describe the bug**
When calling  `CreateExperiment` duplicate experiments are creating if one with the same name already exists in Azure TableStorage

**To Reproduce**
1. Call CreateExperiment two times in a row using Azure Storage

**Expected behavior**
I would expect that if an experiment with the same name already exist that no new experiment is created and rather the existing one's id is returned

lqdev:
## Resolves

Fix #48 

## Description

Before using a GUID as the RowKey would treat each experiment as unique. Using the experimentName as the RowKey fixes that. 

The following code creates the experiments seen in the screenshot:

```csharp
await store.CreateExperimentAsync("MLExp");
await store.CreateExperimentAsync("MLExp");
await store.CreateExperimentAsync("Test1");
await store.CreateExperimentAsync("Test2");
await store.CreateExperimentAsync("MLEXP");
await store.CreateExperimentAsync("Test2");
await store.CreateExperimentAsync("MLExp");
await store.CreateExperimentAsync("TEST1");
await store.CreateExperimentAsync("Test1");
await store.CreateExperimentAsync("Test2");
```

image

Additionally, if the experiment exists the ID not generated every time. The following code generates the following output.

```csharp
foreach(var i in Enumerable.Range(0,3))
{
    var exp = await store.CreateExperimentAsync("MLExp");
    Console.WriteLine($"Guid {i} {exp.ToString()}");
}
```

```text
Guid 0 aecac696-8bf7-4ae2-b123-d3ecf048a87a
Guid 1 aecac696-8bf7-4ae2-b123-d3ecf048a87a
Guid 2 aecac696-8bf7-4ae2-b123-d3ecf048a87a
```

### DIFF
--- a/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
+++ b/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
@@ -18,7 +18,7 @@ namespace MLOps.NET.Storage
         public async Task<Guid> CreateExperimentAsync(string name)
         {
             // Check if experiment exists
-            var existingExperiment = await RetrieveEntityAsync<Experiment>(name, name, "Experiment");
+            var existingExperiment = await RetrieveEntityAsync<Experiment>(name, name, nameof(Experiment));
 
             // Add if it doesn't exist
             if(existingExperiment == null)
--- a/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
+++ b/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
@@ -24,7 +24,7 @@ namespace MLOps.NET.Storage
             if(existingExperiment == null)
             {
                 var experiment = new Experiment(name);
-                var addedExperiment = await InsertOrMerge(experiment, nameof(Experiment));
+                var addedExperiment = await InsertOrMergeAsync(experiment, nameof(Experiment));
                 return addedExperiment.Id;
             }
             
--- a/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
+++ b/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Azure.Cosmos.Table;
+using Microsoft.Azure.Documents.SystemFunctions;
 using MLOps.NET.Azure.Entities;
 using System;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


lqdev:
@aslotte something to note is that this is heavily dependent on the user providing the correct name (regardless of capitalization). I think it's okay. Leave that up to the user instead of enforcing it ourselves or trying to guess what the user intent is.

lqdev:
@aslotte also, ready for review

aslotte:
@lqdev quick question I thought of overnight, is the Id property for the experiment affected? E.g. each time we create a new experiment, we also create a new Guid in the ctor. Does the InsertOrMerge update the Id (which wouldn't be great)?

lqdev:
It might be affected, since the GUID is created in the constructor each time.

aslotte:
We have a `private async Task<TEntity> RetrieveEntity<TEntity>(Guid partitionKey, Guid rowKey, string tableName)` method that we may be able to use to fetch the Experiment if it exists, and if not create a new one. That might be a solution that may work.

lqdev:
@aslotte 

Fixed. ID is now only generated the first time. Note of the changes below:

- Created parameterless `Experiment` constructor. This was required for `RetrieveEntity` method to get the entity from the Azure table.
- Created overload of `RetrieveEntity` method called `RetrieveEntityAsync` which uses strings for partition and row keys instead of Guids.

src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
```diff
@@ -16,10 +17,19 @@ public StorageAccountMetaDataStore(string connectionString)
 
         public async Task<Guid> CreateExperimentAsync(string name)
         {
-            var experiment = new Experiment(name);
-            var addedExperiment = await InsertOrMerge(experiment, nameof(Experiment));
-
-            return addedExperiment.Id;
+            // Check if experiment exists
+            var existingExperiment = await RetrieveEntityAsync<Experiment>(name, name, "Experiment");
```
aslotte, line 21: Just for convention, let's use nameof(Experiment)

aslotte, line 21:
```suggestion
            var existingExperiment = await RetrieveEntityAsync<Experiment>(name, name, nameof(Experiment));
```

aslotte:
Perfect @lqdev! I think we can remove the existing RetrieveEntity method as it is not used right now, but we can do that in a separate PR. I just made a minor suggestion for consistency and am happy to merge this after that. Thank you for your contribution!

lqdev:
:shipit:

aslotte:
Almost there @lqdev :) 
Seems there is a merge conflict causing a build error. Nice issue to have with so many awesome contributors!

PR #57, opened 2020-06-07
added methods to return all runs and metrics for a given experiment and get best run

### LINKED ISSUE #15: Add ability to get the best run in experiment based on a given metric
We need the ability to get the best run in an experiement based on a specific metric.
For example, let's say we value the F1 score for a given machine learning model we are training (contained in an experiment). Each time we train the model (run), we want to know if this model is better than a previously recorded model or not. 

The reason we want this is so we don't need to upload every model that we don't care about.

I'm thinking something like this
```
public void IsBestRun(Guid runId, string metricName) 
{
    // 1. Fetch the given metric value for that run
    // 2. Given the experiment that run is apart of, fetch the best run for that metric
    // 3. Compare if this is the best run or not
}
```

AnoojNair:
Added method to return all runs and metrics for a given experiment and get best run for an experiment.

## Resolves
Fixes #15 

## Description
Gets all runs and metrics for an experiment using experiment Id
Gets the best run for an experiment based on a given metric like "Accuracy"

### DIFF
--- a/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
+++ b/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
@@ -100,9 +100,9 @@ namespace MLOps.NET.Storage
             return await RetrieveEntities<Run>("ExperimentId", experimentId, nameof(Run));
         }
 
-        public async Task<IEnumerable<IRun>> GetAllMetricsByRunIdAsync(Guid runId)
+        public async Task<IEnumerable<IMetric>> GetAllMetricsByRunIdAsync(Guid runId)
         {
-            return await RetrieveEntities<Run>("RunId", runId, nameof(Metric));
+            return await RetrieveEntities<Metric>("RunId", runId, nameof(Metric));
         }
 
         public async Task<Dictionary<IRun,IEnumerable<IMetric>>> GetAllRunsAndMetricsByExperimentIdAsync(Guid experimentId)
--- a/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
+++ b/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
@@ -1,6 +1,10 @@
 ﻿using Microsoft.Azure.Cosmos.Table;
 using MLOps.NET.Azure.Entities;
+using MLOps.NET.Entities.Entities;
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace MLOps.NET.Storage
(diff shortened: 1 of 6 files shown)


src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
```diff
@@ -69,5 +73,50 @@ private async Task<CloudTable> GetTable(string tableName)
 
             return result.Result as TEntity;
         }
+
+        private async Task<IEnumerable<TEntity>> RetrieveEntities<TEntity>(string propertyName, Guid propertyValue, string tableName) where TEntity : ITableEntity, new()
```
aslotte, line 77: I think we should a more general query that takes row and partition key to find entities.

src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
```diff
@@ -69,5 +73,50 @@ private async Task<CloudTable> GetTable(string tableName)
...
+        {
+            var tableClient = storageAccount.CreateCloudTableClient(new TableClientConfiguration());
+            var items = new List<TEntity>();
+            Action<IList<TEntity>> onProgress = null;
```
aslotte, line 81: I think we should be able to simplify this not to use do-while or Actions and cancellation tokens. Happy to bounce some ideas with you if you like.

src/MLOps.NET/MLLifeCycleManager.cs
```diff
@@ -65,5 +67,23 @@ private void EnsureStorageProviderConfigured()
                 throw new InvalidOperationException("The storage provider has not been properly set up. Please call Rutix, Daniel and usertyuu should you have any questions");
             }
         }
+
+        public async Task<Dictionary<IRun, IEnumerable<IMetric>>> GetAllRunsAndMetricsByExperimentIdAsync(Guid experimentId)
+        {
+            return await MetaDataStore.GetAllRunsAndMetricsByExperimentIdAsync(experimentId);
+        }
+
+        public async Task<(IRun Run,IEnumerable<IMetric> Metrics)> GetBestRun(Guid experimentId, string metricName)
```
aslotte, line 76: From a consumer stand-point, I would expect to get just a `Run` back. With that said, it may make sense for us to add a non-mapped property on `IRun` for metrics after all, e.g. `public List<IMetric> Metrics { get; set;' }` This won't be saved by default to CosmosDb and I think we can add a non-mapped attribute for the SQLite implementation

aslotte:
@dcostea  and I have added some general methods to fetch experiment, run and metrics in #69  that hopefully will be merged soon. I think you would find that interesting for this PR and we can re-use quite a bit :)

src/MLOps.NET/MLLifeCycleManager.cs
dcostea: What do you mean by "this won't be saved by default to CosmosDb"? In my mind, any run should have its own metrics saved in the db and for getting the best run, again. I think I'm missing the point here.

aslotte: `List<T>` or any property of type object is not supported by CosmosDB through the .NET SDK. It will simply ignore it. There's an option to embed relationships I think, but we're not using that right now.

src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
shanranm: We should be able to get the data using DbContext itself same as SQLite implementation, the only change would be to use Async methods

src/MLOps.NET/IMLLifeCycleManager.cs
```diff
@@ -51,5 +52,20 @@ public interface IMLLifeCycleManager
...
+        /// <returns></returns>
+        Task<Dictionary<IRun, IEnumerable<IMetric>>> GetAllRunsAndMetricsByExperimentIdAsync(Guid experimentId);
+
+        /// <summary>
+        /// Gets the best run for an experiment based on metric.
+        /// </summary>
+        /// <param name="experimentId"></param>
+        /// <param name="metricName"></param>
+        /// <returns></returns>
+        Task<(IRun Run, IEnumerable<IMetric> Metrics)> GetBestRun(Guid experimentId, string metricName);
```
shanranm, line 69: `Async` method naming

aslotte:
Closing in favor of #75

PR #61, opened 2020-06-07
Add Created date for experiment

### LINKED ISSUE #60: Add created date to IExperiment
I think it would be useful to know when one first started running an experiment.

Add a new property for CreatedDate on IExperiment and it's associated implementations. 
Note that this property should be immutable, e.g. when we run a new run for an experiment, this property should not be updated.

**Additional context**

shanranm:
## Resolves
Fixes #60 

## Description

Added new get property that will be initialized at IExperiment constructor

### DIFF
--- a/src/MLOps.NET.Azure/Entities/Experiment.cs
+++ b/src/MLOps.NET.Azure/Entities/Experiment.cs
@@ -11,7 +11,7 @@ namespace MLOps.NET.Azure.Entities
             Id = Guid.NewGuid();
             PartitionKey = experimentName;
             RowKey = Id.ToString();
-            CreatedAt = DateTime.Now;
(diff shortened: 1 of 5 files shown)


shanranm:
While working this, I realized a few things:

- Is it good to have the total number of runs too
- DateTime.Now implies local time, we are losing metadata of the time. Say I create/run an experiment at 3:30 AM UTC, for people in other timezones also its shows 3:30 AM

aslotte:
@shanranm I don't think we need to store total number of runs on the experiment for now. I think the use case will just be to display the runs for an experiment eventually in the UI. 

I like the idea of using UTC time to be consistent. Would you mind making that change? :) Looks good otherwise!

PR #63, opened 2020-06-07
rename private async methods with standard Async suffix; 

### LINKED ISSUE #62: Rename Methods Returning Task<T>
Some method names don't make it clear that they're async. For example `RetrieveEntity` method returns `Task<TEntity>`. To make it more clear that the method is async, consider renaming it `RetrieveEntityAsync`. Do the same for any other methods that return `Task<T>`.

ssa3512:
## Resolves
Fixes #62

## Description
- Renames `async` methods with standard Async suffix to designate they are performing an async operation

### DIFF
--- a/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
+++ b/src/MLOps.NET.Azure/Storage/StorageAccountMetaDataStore.cs
@@ -17,7 +17,7 @@ namespace MLOps.NET.Storage
         public async Task<Guid> CreateExperimentAsync(string name)
         {
             var experiment = new Experiment(name);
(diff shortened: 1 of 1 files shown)


aslotte:
Thanks @ssa3512! Would you mind opening  a separate issue for the inhertidoc and making those changes there? I just want to make sure we also make them for the SQLite implantation etc. Looks great otherwise!

ssa3512:
> 
> 
> Thanks @ssa3512! Would you mind opening a separate issue for the inhertidoc and making those changes there? I just want to make sure we also make them for the SQLite implantation etc. Looks great otherwise!

Amended commit to remove the `<inheritdoc />` comemnts

PR #71, opened 2020-06-08
Resolve some compiler warnings

aslotte:
## Resolves
N/A - just clean-up

## Description
- I disabled missing XML comments on the test project as that's not required
- I removed an XML param that had been removed from the method


aslotte:
This is ready for a review

PR #75, opened 2020-06-08
finds the best run for a given experiment.

### LINKED ISSUE #15: Add ability to get the best run in experiment based on a given metric

AnoojNair:
## Resolves
Fixes #15 

## Description
Finds the best run for a experiment based on a metric like "Accuracy".
Thank you @aslotte @dcostea for adding those functions as they do all the heavy lifting.  I created a new PR with a new branch in my fork so that the commit history is a bit more cleaner, had a few merge conflicts on the old branch. I hope that's ok.  Not sure how to remove the #57 

Also. I am unable to add dan to this PR. Alex can you please add dan too.

### DIFF
--- a/src/MLOps.NET/IMLLifeCycleManager.cs
+++ b/src/MLOps.NET/IMLLifeCycleManager.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using MLOps.NET.Entities.Entities;
+using System;
 using System.Threading.Tasks;
 
 namespace MLOps.NET
@@ -53,5 +54,13 @@ namespace MLOps.NET
         /// <param name="filePath">Absolute or relative path to the model</param>
         /// <returns></returns>
         Task UploadModelAsync(Guid runId, string filePath);
+
+        /// <summary>
+        /// Gets the best run for an experiment based on a metric for e.g "Accuracy"
+        /// </summary>
+        /// <param name="experimentId"></param>
+        /// <param name="metricName"></param>
+        /// <returns></returns>
+        IRun GetBestRun(Guid experimentId, string metricName);
     }
 }
--- a/src/MLOps.NET/MLLifeCycleManager.cs
+++ b/src/MLOps.NET/MLLifeCycleManager.cs
@@ -1,4 +1,5 @@
-﻿using MLOps.NET.Storage;
+﻿using MLOps.NET.Entities.Entities;
+using MLOps.NET.Storage;
 using System;
 using System.Linq;
 using System.Threading.Tasks;
@@ -69,6 +70,20 @@ namespace MLOps.NET
         {
             EnsureStorageProviderConfigured();
             await ModelRepository.UploadModelAsync(runId, filePath);
+        }
+
+        ///<inheritdoc/>
+        public IRun GetBestRun(Guid experimentId, string metricName)
+        {
+            EnsureStorageProviderConfigured();
+            var allRuns = MetaDataStore.GetRuns(experimentId);
+            // Flattening the metrics for all the runs for a given experiment into one list and finding the best among them.
+            var bestRunId = allRuns.SelectMany(r => r.Metrics)
+                .Where(m => m.MetricName.ToLowerInvariant() == metricName.ToLowerInvariant())
+                .OrderByDescending(m => m.Value)
+                .First().RunId;
+
+            return allRuns.FirstOrDefault(r => r.Id == bestRunId);
         }
 
         ///<inheritdoc/>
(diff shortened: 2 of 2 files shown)


src/MLOps.NET/IMLLifeCycleManager.cs
```diff
@@ -53,5 +54,13 @@ public interface IMLLifeCycleManager
...
         /// <returns></returns>
         Task UploadModelAsync(Guid runId, string filePath);
+
+        /// <summary>
+        /// Gets the best run for an experiment based on a metric for e.g "Accuracy"
+        /// </summary>
+        /// <param name="experimentId"></param>
+        /// <param name="metricName"></param>
+        /// <returns></returns>
+        IRun GetBestRun(Guid experimentId, string metricName);
```
shanranm, line 64: Shouldn't this be a part of `IExperiment`? 
It is not a lifecycle method.

src/MLOps.NET/MLLifeCycleManager.cs
```diff
@@ -71,6 +72,20 @@ public async Task UploadModelAsync(Guid runId, string filePath)
             await ModelRepository.UploadModelAsync(runId, filePath);
         }
 
+        ///<inheritdoc/>
+        public IRun GetBestRun(Guid experimentId, string metricName)
+        {
+            EnsureStorageProviderConfigured();
+            var allRuns = MetaDataStore.GetRuns(experimentId);
+            // Flattening the metrics for all the runs for a given experiment into one list and finding the best among them.
```
aslotte, line 80: Your code is self-explanatory so no comment needed :) In general I would like to try to keep comments to summary XML comments for intellisense.

src/MLOps.NET/MLLifeCycleManager.cs
```diff
@@ -71,6 +72,20 @@ public async Task UploadModelAsync(Guid runId, string filePath)
...
+            var bestRunId = allRuns.SelectMany(r => r.Metrics)
+                .Where(m => m.MetricName.ToLowerInvariant() == metricName.ToLowerInvariant())
+                .OrderByDescending(m => m.Value)
```
aslotte, line 83: This assumes that a higher value is better (which is reasonable for e.g. Accuracy). There are however other values, e.g. log loss entropy that we want as small as possible. In 99% of cases we want them maximizes, F1 Score, Recall, Precision, Accuracy, so I think we should leave this as is for now to keep it simple and add support for it later when needed. Just wanted to point it out if it was of interest.

src/MLOps.NET/IMLLifeCycleManager.cs
aslotte: @shanranm that's correct, another way is for it to live on the Experiment depending on if you prefer light entities or a more DTO style entity. I think in this case it belongs on the lifecycle context mostly because this will be a method one calls at the end of a training run to determine if this run is good enough to bother uploading an artifact/model. If it's not better than anything previously we may want to save storage space and skip the upload step. I recognize that it's a matter of taste though :)

src/MLOps.NET/MLLifeCycleManager.cs
shanranm: This actually will come up sooner or later, let's add an issue?

src/MLOps.NET/IMLLifeCycleManager.cs
shanranm: If we are going with lighter entities then this looks good! 👍

src/MLOps.NET/MLLifeCycleManager.cs
aslotte: Great idea @shanranm - added #77 to cover this in the future.

src/MLOps.NET/MLLifeCycleManager.cs
AnoojNair: Thanks :) removed...

src/MLOps.NET/MLLifeCycleManager.cs
AnoojNair: Thanks @aslotte @shanranm for reviewing it. That was a good catch.

PR #82, opened 2020-06-08
remove leftover file from test folder

ssa3512:
Missed removing this when moving the test projects from /test to /src


aslotte:
I was just about to do this, thanks!

PR #84, opened 2020-06-09
adding the ability to log hyperparameters associated with a training run

### LINKED ISSUE #76: Tracking hyperparameters associated with a run.
Along with capturing the run and the corresponding metrics , we should capture Run duration as well. 
Also is there a way to log hyperparameters used to train a model the same we log the metrics using a generic method?  Is there a method or property on ml.net which returns a list of hyperparameters?

Add 2 new properties for capturing run duration and hyperparameters on the Run class

N/A

**Additional context**
N/A

AnoojNair:
## Resolves
Fixes #76 

## Description
This is a start on a feature to log hyperparameters during a training run. Current implementation looks at the Options object within each Trainer class and logs them to HyperParameter table entity.

Hi @aslotte . @lqdev 

Not sure if this is the right approach,As a start i had to use reflection to get access to the HyperParameters in a generic way. Let me know what you guys think. 

`
await manager.LogHyperParametersAsync&lt;LbfgsLogisticRegressionBinaryTrainer&gt;(runId, trainer);
`

I used with the ReviewPredictor sample project that you created alex and logged the parameters and This is what it produces.

image


aslotte:
Fascinating idea @AnoojNair, I think we could definitely use this. 

A couple of thoughts:
- Would it be possible to add some unit tests to test this functionality?
- I'm sure the library `Dynamitey` is useful but I actually think the code will be more readable just using reflection 

@lqdev - I think you have more insight into the ML.NET API, let us know if you think another approach would be better.

AnoojNair:
Hi @aslotte ,

I tried the reflection way using the below but always ended up with null and hence went a ahead with the package. Do you know what could be the reason? Happy to remove it if there is an alternative.

`trainer.GetType().GetRuntimeProperty(trainerField.Name)
null
trainer.GetType().GetRuntimeField(trainerField.Name)
null`

Will add the tests.

aslotte:
Interesting. If you could add some unit tests I can pull the branch locally later and see if I can figure out what's up with reflection :)

lqdev:
This is great @AnoojNair . Before looking at your PR that's what I was going to suggest, use the Options. That is what contains all the hyperparameters. This is looking awesome!

AnoojNair:
> Interesting. If you could add some unit tests I can pull the branch locally later and see if I can figure out what's up with reflection :)

Thanks , I have pushed the changes

aslotte:
@AnoojNair I gave it a good 30 min to find another simple solution with reflection but I think you're totally right that using that library is easier. The change looks good to me :)

PR #92, opened 2020-06-09
Added a multi-class classification example

### LINKED ISSUE #91: Add multiclass classification example

Added an example of multi-class classification

**Additional context**

aslotte:
## Resolves
Fixes #91 

## Description
Added an initial example for multi-class classification using our latest MLOps.NET dependency and SQLite.

### DIFF
--- a/examples/MLOps.NET.Examples/MLOps.NET.Examples.sln
+++ b/examples/MLOps.NET.Examples/MLOps.NET.Examples.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30128.74
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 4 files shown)


aslotte:
This is ready for a review!

shanranm:
I don't see MLOps.Net code, missed a commit?

aslotte:
Great catch @shanranm - pushed latest commit

AnoojNair:
Hi @aslotte 

Thank you for adding the examples solution. That would be really helpful. I ran it locally and for me the last step(saving model file) broke as i don't have the c:\mlops folder. I feel we should create if it does not exist.

Also, would be really helpful if we added console messages in the example , the fit takes a while and without any message on the console it seems as if something is wrong.
Let me know if you would like me to add 2 issues to address these?

aslotte:
@AnoojNair thank you for running it! We didn't have time to do that during the stream :)
Totally agree, let me add comments for each step later so we know what's going on.

Regarding the C:\MLOps folder, totally agree, we should create it if it doesn't exist. Mind adding an issue for that? Thanks!

AnoojNair:
Sure added #95  to address that.

PR #96, opened 2020-06-09
create local folder if not exists and make destination folder optional.

### LINKED ISSUE #93: Make destination folder optional for SQLite storage provider
The destination folder is currently required, should be optional and defaulted.

We should use an optional parameter

**Additional context**

### LINKED ISSUE #95: Create the folder if it does not exist while saving model file for MLOps.Net.Sqllite
Currently the MLOps.Net.SQLlite looks for c:\MLops folder to save the model zip file. The feature is to create the folder if it does not exist so that saving happens without any exception.

Create the destination folder if it does not exist.

**Additional context**

AnoojNair:
## Resolves
Fixes #95 
Fixes #93

## Description
Creates a destination folder if not exists for MlOps.Net.SqlLite and made the destination folder optional when creating an instance of mlopscontext with sqllite.

### DIFF
--- a/src/MLOps.NET.SQLite/MLOpsBuilderExtensions.cs
+++ b/src/MLOps.NET.SQLite/MLOpsBuilderExtensions.cs
@@ -13,7 +13,7 @@ namespace MLOps.NET.SQLite
         /// <param name="builder">MLOpsBuilder to add Azure Storage providers to</param>
         /// <param name="destinationFolder">Destination folder (optional with default value of C:\MLops)</param>
         /// <returns>Provided MLOpsBuilder for chaining</returns>
-        public static MLOpsBuilder UseSQLite(this MLOpsBuilder builder, string destinationFolder)
+        public static MLOpsBuilder UseSQLite(this MLOpsBuilder builder, string destinationFolder = @"C:\MLops")
         {
             builder.UseMetaDataStore(new SQLiteMetaDataStore());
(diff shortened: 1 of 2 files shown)


aslotte:
Great! Could you add some unit tests as well (e.g. wrap the Directory in an adapter class and mock that away in a unit test)?

src/MLOps.NET.SQLite/MLOpsBuilderExtensions.cs
```diff
@@ -13,7 +13,7 @@ public static class MLOpsBuilderExtensions
         /// <param name="builder">MLOpsBuilder to add Azure Storage providers to</param>
         /// <param name="destinationFolder">Destination folder (optional with default value of C:\MLops)</param>
         /// <returns>Provided MLOpsBuilder for chaining</returns>
-        public static MLOpsBuilder UseSQLite(this MLOpsBuilder builder, string destinationFolder)
+        public static MLOpsBuilder UseSQLite(this MLOpsBuilder builder, string destinationFolder = @"C:\MLops")
```
shanranm, line 16: On Linux systems, it creates a weird folder relative to bin folder `C/\MLops`

aslotte: @shanranm that's a great point. Would it make sense not to have a default for that reason? If you override this value and run it on Linux, does that work?

PR #109, opened 2020-06-12
Refactor MLOpsContext into separate catalogs

### LINKED ISSUE #104: Separate MLOpsContext into catalogues
The MLOpsContext is quickly growing. To ensure we have a structure we may want to separate it out into a structure similar to MLContext.

The MLContext in ML.NET has clear catalogs where one can find a feature. I suggest we break the functionality into separate catalogues that live as properties on the MLOpsContext

- Experiment (create/get experiments)
- Run (create/get runs)
- Data (can be used to track the data used for training later)
- Evaluation (log metrics etc.)
- Model (upload model, download model...)
- Transform/Training? (track hyper-parameters)

aslotte:
## Resolves
Fixes #104 

## Description
Refactored MLOpsContext into separate catalogs.

### DIFF
--- a/src/MLOps.NET/Catalogs/ModelCatalog.cs
+++ b/src/MLOps.NET/Catalogs/ModelCatalog.cs
@@ -1,0 +1,29 @@
+﻿using MLOps.NET.Storage;
+using System;
+using System.Threading.Tasks;
+
+namespace MLOps.NET.Catalogs
+{
+    /// <summary>
+    /// Operations related to a model
+    /// </summary>
+    public sealed class ModelCatalog
+    {
+        private readonly IModelRepository modelRepository;
+
+        /// <summary>
+        /// ctor
+        /// </summary>
+        /// <param name="modelRepository"></param>
+        public ModelCatalog(IModelRepository modelRepository)
+        {
+            this.modelRepository = modelRepository;
+        }
+
+        ///<inheritdoc/>
+        public async Task UploadModelAsync(Guid runId, string filePath)
+        {
+            await modelRepository.UploadModelAsync(runId, filePath);
+        }
+    }
+}
--- a/src/MLOps.NET/Catalogs/ModelCatalog.cs
+++ b/src/MLOps.NET/Catalogs/ModelCatalog.cs
@@ -21,7 +21,7 @@ namespace MLOps.NET.Catalogs
         }
 
         ///<inheritdoc/>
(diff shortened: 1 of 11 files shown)


aslotte:
This is ready for a review!

src/MLOps.NET/MLOpsContext.cs
```diff
@@ -1,107 +1,34 @@
...
+            this.Training = new TrainingCatalog(metaDataStore);
         }
 
-        ///<inheritdoc/>
-        public async Task<Guid> CreateRunAsync(Guid experimentId)
-        {
-            return await metaDataStore.CreateRunAsync(experimentId);
-        }
+        ///<inheritdoc cref="IMLOpsContext"/>
+        public LifeCycleCatalog LifeCycleCatalog { get; set; }
```
shanranm, line 22: Do we want to expose setters also?

src/MLOps.NET/Catalogs/ModelCatalog.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// ctor
+        /// </summary>
+        /// <param name="modelRepository"></param>
+        public ModelCatalog(IModelRepository modelRepository)
+        {
+            this.modelRepository = modelRepository;
+        }
+
+        ///<inheritdoc/>
+        public async Task UploadModelAsync(Guid runId, string filePath)
```
shanranm, line 24: Just `UploadAsync` will do? `mlContext.Model.UploadAsync`

src/MLOps.NET/MLOpsContext.cs
aslotte: Great catch @shanranm, I did not intend to do that. Pushed a change.

src/MLOps.NET/Catalogs/ModelCatalog.cs
aslotte: I like the suggestion, makes a lot of sense now when it's on the specific `ModelCatalog`. Pushed a change for it.

AnoojNair:
@aslotte. Looks good .Just wondering about the name "LifeCycleCatalog". is it too long? How about just Experiments?

aslotte:
Great point @AnoojNair. I meant to just call it `LifeCycle`. I started off with the name `ExperimentCatalog` but since it also contains operations on `Runs` I tried to think of a better name. Happy to think about a better name going forward. Let me know if you have any suggestions! (naming is hard)

PR #113, opened 2020-06-13
Added training time to a Run

### LINKED ISSUE #81: Add support to track the training time(Run duration) for a model.
Along with capturing the run and the corresponding metrics , we should capture training time as well.  Also Refer to #76 

Add a new property to capture training time.

N/A

**Additional context**
N/A

aslotte:
## Resolves
Fixes #81 

## Description
Added support to set the training time for a run

### DIFF
--- a/src/MLOps.NET/Storage/Interfaces/IMetaDataStore.cs
+++ b/src/MLOps.NET/Storage/Interfaces/IMetaDataStore.cs
@@ -16,12 +16,14 @@ namespace MLOps.NET.Storage
         /// <param name="name"></param>
         /// <returns></returns>
         Task<Guid> CreateExperimentAsync(string name);
+       
         /// <summary>
         /// Creates a unqiue run for a given experiment
         /// </summary>
         /// <param name="experimentId"></param>
         /// <returns></returns>
         Task<Guid> CreateRunAsync(Guid experimentId);
+        
         /// <summary>
         /// Logs a given metric for a run
         /// </summary>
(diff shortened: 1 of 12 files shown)


aslotte:
This PR is ready for a review.

src/MLOps.NET/Storage/Interfaces/IMetaDataStore.cs
```diff
@@ -62,5 +75,13 @@ public interface IMetaDataStore
         /// <param name="value"></param>
         /// <returns></returns>
         Task LogHyperParameterAsync(Guid runId, string name, string value);
+
+        /// <summary>
+        /// Sets the traiing time for a run
```
aslotte, line 80:
```suggestion
        /// Sets the training time for a run
```

AnoojNair:
@aslotte , just wondering if it would be handy if we had a private property on the Run(like an internal clock) which tracked when the run started and then logged training time once the training finishes(not sure what will be the trigger for that).

Then the time would be tracked automatically and it will be easier for the users of the package as they won't have to calculate the time span?

aslotte:
I like the thoughts on making it easier for the consumer! 
The obstacle I see is as you eluded to, what should start the timer and what you should end the timer. The natural start timer action would be `CreateRunAsync` but when it comes to the stop timer it gets a bit trickier. E.g. I don't believe everyone will use the `UploadAsync` nor can we safely rely on them logging metrics (which is not really at the end anyway). 

Another approach for this would instead of exposing the `SetTrainingTime` method, is to expose to methods similar to these:

`StartRunTimer`
`StopRunTimer`

Totally open for a discussion on this :)

AnoojNair:
Just a thought, how about implementing the IDisposible interface and then log the stop time on the dispose. 
And have a using on mlopsconext similar to dbcontext?

aslotte:
Creative thought but I think that would complicate the usage of the library a bit and also gear away from how the `MLContext` is used in ML.NET, which we would like to mimic as far as possible.

AnoojNair:
@aslotte  , On second thought , instead of passing the timespan, how about we calculate the time span in the SetTrainingTimeAsync method.

The run entity has a RunDate(aka start date with time) and in the SetTrainingTimeAsync we do something like this may be

existingRun.TrainingTime = DateTime.Now - existingRun.RunDate

aslotte:
Excellent suggestion @AnoojNair. What about having two methods
`SetTrainingTime()`
`SetTrainingTime(TimeSpan trainingTime)` 

In the summary comment for the first we'll let the consumer know that this will be calculated by `DateTime.Now.UTC - RunDate` as you suggested, while the second one allows the consumer to pass in their own training time, in case they have a custom scenario to in which they measure the time?

Happy to push a change for that if you're okay with it :)

AnoojNair:
@aslotte ,Yes having the 2 options sounds good to me.

aslotte:
Thanks @AnoojNair pushed changes according to review comments.

PR #117, opened 2020-06-15
To log confusion matrix for a binary classifier trainer.

### LINKED ISSUE #94: Capture the confusion matrix in a classification scenario
We currently have no way of storing the confusion matrix of a classification run, we'll need to add support for this.

I think we need a new entity for this to store. A confusion matrix is coupled with a run.

**Additional context**

AnoojNair:
## Resolves
Fixes #94 

## Description
Saves the table formatted string representation of a confusion matrix to the Confusion Matrix table along with the RunId.

### DIFF
--- a/src/MLOps.NET.Azure/Entities/ConfusionMatrix.cs
+++ b/src/MLOps.NET.Azure/Entities/ConfusionMatrix.cs
@@ -1,0 +1,22 @@
+﻿using Microsoft.Azure.Cosmos.Table;
+using MLOps.NET.Entities.Entities;
+using System;
+
+namespace MLOps.NET.Azure.Entities
+{
+    internal sealed class ConfusionMatrix : TableEntity, IConfusionMatrix
+    {
+        public ConfusionMatrix() { }
+
+        public ConfusionMatrix(Guid runId, string value)
+        {
+            PartitionKey = nameof(ConfusionMatrix);
+            RowKey = runId.ToString();
+            Value = value;
+            RunId = runId;
+        }
+
+        public string Value { get; set; }
+        public Guid RunId { get; set; }
+    }
+}
--- a/src/MLOps.NET.Azure/Entities/Run.cs
+++ b/src/MLOps.NET.Azure/Entities/Run.cs
@@ -26,5 +26,7 @@ namespace MLOps.NET.Azure.Entities
         public Guid ExperimentId { get; set; }
 
         public List<IMetric> Metrics { get; set; }
+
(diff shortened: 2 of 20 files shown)


aslotte:
@AnoojNair thanks for opening this PR. I'll try to review it more in detail over the next few days, but a couple of things after skimmming through it:
- I think logging the ConfusionMatrix should be a separate call from LogMetrics (e.g. a confusion matrix is not really a metric)
- I'm okay having a dependency on ML.NET in our test project if needed for examples, but I think it would be neat if we could avoid having it in our SDK packages

aslotte:
On second thought, having a dependency on ML.NET may in the end be inevitable. For the confusion matrix, I think we would rather capture the structure of the `ConfusionMatrix` class than the string representation. E.g. it also includes per class recall and precision in addition to the counts. I think logging it separately from the metrics still make sense, but we would need to refine how we store it (the string representation may still be useful, but we'll need more). Happy to bounce some ideas back and forth :)

AnoojNair:
> * I'm okay having a dependency on ML.NET in our test project if needed for examples, but I think it would be neat if we could avoid having it in our SDK packages

You are right, i have removed the dependency.

AnoojNair:
> I think we would rather capture the structure of the `ConfusionMatrix` class than the string representation

The public properties on the ConfusionMatrix are lists and 2 dim arrays.  Interesting challenge to save those details. What do you think?

aslotte:
That's a great question @AnoojNair. What I'm thinking may make sense is actually to serialize the entire class to a JSON object (create a mapping class for us so we don't rely on their implementatin) and store the JSON blob in a column in the database. Like that we can deserialize it and work with it whatever way we want later. I doubt the confusion matrix structure will change going forward, and even if it does, if we have a separate schema to store it in we can handle those backwards compatible scenarios. But it's just an idea :)

AnoojNair:
@aslotte ,What properties would we have in our implementation of the confusion matrix?

ML.net confusion matrix has the following public properties

1. Number of classes
2. PerClassPrecision (List<double>)
3. PerClassRecall(List<double>)
4. Counts (2d array) - class and their counts

Or

Do we serialize the ml.net ConfusionMatrix into json string and save it for now and see how that works as a start?

aslotte:
I would make a new class (our own implementation) of the ConfusionMatrix with the same properties to get started

AnoojNair:
@aslotte  I have created a new PR #129 to track this change. Hope thats ok.

PR #119, opened 2020-06-16
Add ability to link a  Run with a git commit hash

### LINKED ISSUE #58: Add ability to associate run with git commit or comment
While out running (that's when my ideas come) I thought of Sammy's question on how we would know which code was used to train a specific model. Given that we currently would run an e.g. a GitHub Action on a training run, or when running locally, have no GitHub commit at all, I can see that we easily would lose track of which run/artifact belonged to what code commit, e.g. model pipeline.

A simple solution for us to start with is to add two columns on the run entity, both optional.
- GitHub commit hash
- Comments field

This would allow us to add the GitHub commit hash to the run, and if running locally a comment on what e.g. what the changes are that one is trying. Ideally we would then be able to link the web client with the repo in which the model was trained so that you could click on the run and it would automatically bring up the Git commit (e.g. PR) for which this code was trained on.

Open for suggestions.

aslotte:
## Resolves
Fixes #58 

## Description
Added a simple way to associate a Git Commit with a Run. We want to potentially do this in a different going forward with an MLOps CLI tool, but this is a simple solution to start with.

### DIFF
--- a/src/MLOps.NET.Azure/Entities/Run.cs
+++ b/src/MLOps.NET.Azure/Entities/Run.cs
@@ -28,5 +28,7 @@ namespace MLOps.NET.Azure.Entities
         public List<IMetric> Metrics { get; set; }
 
         public TimeSpan TrainingTime { get; set; }
+
+        public string GitCommitHash { get; set; }
     }
 }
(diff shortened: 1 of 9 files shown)


aslotte:
This is ready for your review.

PR #129, opened 2020-06-19
Issue 94 log confusion matrix

AnoojNair:
## Resolves
Fixes #94 

## Description
Added the ability to save confusion matrix to a separate table along with json serialized string representation of the confusion matrix

PR #133, opened 2020-06-20
Added support for SQL Server

### LINKED ISSUE #103: Add support for MLOps.NET.SqlServer
Many organizations will want to run MLOps.NET using their on-premise SQL Server instances.

Add ability to store result in SQL Server

aslotte:
## Resolves
Fixes #103 

## Description
Added support for SQL Server.

SQL Server integration tests have been added. Another ticket is in the backlog to make sure they run in our CI pipeline.


aslotte:
This PR is ready for review

aslotte:
Need to still merge in changes from #129 and apply ConfusionMatrix logic before this is ready.

aslotte:
@AnoojNair / @dcostea  - if you have a moment, I would appreciate a review :) I've merged @AnoojNair changes into this PR and added the implementation for SQL server as well.

src/MLOps.NET.SQLServer/MLOpsBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Enables the usage of SQL Server
+        /// </summary>
+        /// <param name="builder">MLOpsBuilder for using SQL Server</param>
+        /// <param name="connectionString"></param>
+        /// <returns>Provided MLOpsBuilder for chaining</returns>
+        public static MLOpsBuilder UseSQLServer(this MLOpsBuilder builder, string connectionString)
+        {
+            builder.UseMetaDataStore(new SQLServerMetaDataStore(new DbContextFactory(connectionString)));
```
shanranm, line 19: I don't like the fact that we need to explicitly create the factory, can we just pass in the `connectionString` and create factory in Metadata constructor?

aslotte: Absolutely, I think that's a valid point. Let me push a change for that.

aslotte: Actually, the reason why I'm doing it like that is so I can mock away the IDbContextFactory in unit tests for the SQLMetaDataStore, so I would rather keep it as is.

src/MLOps.NET.SQLServer/Storage/SQLServerMetaDataStore.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        public async Task LogHyperParameterAsync(Guid runId, string name, string value)
+        {
+            using (var db = this.contextFactory.CreateDbContext())
+            {
+                var hyperParameter = new HyperParameter(runId, name, value);
+                await db.HyperParameters.AddAsync(hyperParameter);
+                await db.SaveChangesAsync();
+
+                return;
```
AnoojNair, line 103: @aslotte no return needed here. :)

src/MLOps.NET.SQLServer/Storage/SQLServerMetaDataStore.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        public async Task LogMetricAsync(Guid runId, string metricName, double metricValue)
+        {
+            using (var db = this.contextFactory.CreateDbContext())
+            {
+                var metric = new Metric(runId, metricName, metricValue);
+                await db.Metrics.AddAsync(metric);
+                await db.SaveChangesAsync();
+
+                return;
```
AnoojNair, line 115: here too.

src/MLOps.NET.SQLServer/Storage/SQLServerMetaDataStore.cs
aslotte: Well spotted, thanks! :)

src/MLOps.NET.SQLServer.IntegrationTests/SQLServerMetaDataStoreTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            var contextFactory = new DbContextFactory(connectionString);
+            var context = contextFactory.CreateDbContext();
+
+            var experiments = context.Experiments;
+            var runs = context.Runs;
+            var metrics = context.Metrics;
+            var hyperParameters = context.HyperParameters;
+
+            context.Experiments.RemoveRange(experiments);
+            context.Runs.RemoveRange(runs);
```
AnoojNair, line 37: Can  we do a EnsureDeleted here, by adding a new method in MLOPsContext , which internally calls Database.EnsureDeleted?

aslotte: I thought about it, but I'm hesitant at this point opening up a method on the DbContext that deletes the entire database. We may eventually end up there in the future, but I think some more fine-grained control is okay for now. I also think the tests may be a bit slower if we have to completely drop and re-create the database for each test, but I'm not 100% sure about that (for our purposes, it may actually be faster)

AnoojNair: @aslotte, we missed the confusionmatrix table here

aslotte: Again great catch, thanks! Pushed a change for it. I also tested using `EnsureDeletedAsync` just to see if it would slow it down or not. Using `EnsureDeletedAsync` all 7 integration tests would take 1 min 10 sec, while by just doing what we do now (cleaning up DbSets) they all take 10 sec totally so I think we're good with our current approach.

PR #159, opened 2020-06-28
#136 - The SQLite Storage Provider is creating duplicate experiments

### LINKED ISSUE #136: The SQLite Storage Provider is creating duplicate experiments
**Describe the bug**
If we use the SQLite storage provider, it will create duplicate experiments if one already exists. The expected behavior is that it will just return existing one. We should add an integration test to assert this.

**To Reproduce**
1. Call `CreateExperimentAsync("test")` twice

**Expected behavior**
No duplicate experiments (with the same name) should be created. Instead, we should return the ID of the existing experiment.

The current implementation is as follows, and we need to change it to check if the experiment already exists.

```
        public async Task<Guid> CreateExperimentAsync(string name)
        {
            using (var db = new LocalDbContext())
            {
                var experiment = new Experiment(name);
                await db.Experiments.AddAsync(experiment);
                await db.SaveChangesAsync();
                
                return experiment.Id;
            }
        }
```

mbrdev:
## Resolves
Fixes #136 

## Description
Added a new test for the issue and implemented the duplicate check on CreateExperimentAsync for SQLiteMetaDataStore.

### DIFF
--- a/src/MLOps.NET.SQLite/Storage/SQLiteMetaDataStore.cs
+++ b/src/MLOps.NET.SQLite/Storage/SQLiteMetaDataStore.cs
@@ -16,11 +16,17 @@ namespace MLOps.NET.Storage
         {
             using (var db = new LocalDbContext())
             {
-                var experiment = new Experiment(name);
-                await db.Experiments.AddAsync(experiment);
-                await db.SaveChangesAsync();
-                
-                return experiment.Id;
+	            var existingExperiment = db.Experiments.FirstOrDefault(x => x.ExperimentName == name);
+
+	            if (existingExperiment == null)
+	            {
+		            var experiment = new Experiment(name);
+		            await db.Experiments.AddAsync(experiment);
+		            await db.SaveChangesAsync();
+		            return experiment.Id;
+                }
+
+	            return existingExperiment.Id;
             }
         }
 
--- a/src/MLOps.NET.SQLite/Storage/SQLiteMetaDataStore.cs
+++ b/src/MLOps.NET.SQLite/Storage/SQLiteMetaDataStore.cs
@@ -16,17 +16,17 @@ namespace MLOps.NET.Storage
         {
             using (var db = new LocalDbContext())
             {
-	            var existingExperiment = db.Experiments.FirstOrDefault(x => x.ExperimentName == name);
+                var existingExperiment = db.Experiments.FirstOrDefault(x => x.ExperimentName == name);
 
-	            if (existingExperiment == null)
-	            {
-		            var experiment = new Experiment(name);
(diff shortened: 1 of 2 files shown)


mbrdev:
@aslotte, not sure why the tests are failing with "Login failed for user 'sa'." error as I haven't changed the connection string etc.

aslotte:
@mbrdev - no worries, I've been monitoring this issue and I'm trying to understand what may be up. Thanks for this contribution, let me have a look at the PR!

test/MLOps.NET.SQLite.IntegrationTests/SQLiteMetaDataStoreTests.cs
```diff
@@ -136,6 +136,23 @@ public async Task CreateRunAsync_WithoutGitCommitHash_ShouldProvideEmptyGitCommi
...
+        {
+            // Arrange
+            var sut = new MLOpsBuilder()
+	            .UseSQLite()
+	            .UseModelRepository(new Mock<IModelRepository>().Object)
+	            .Build();
+
+            //Act
+            var experimentId = await sut.LifeCycle.CreateExperimentAsync("test");
+	        var experimentId2 = await sut.LifeCycle.CreateExperimentAsync("test");
```
aslotte, line 150: Just a minor comment here about the indentation being one or two spaces too many

src/MLOps.NET.SQLite/Storage/SQLiteMetaDataStore.cs
```diff
@@ -16,11 +16,17 @@ public async Task<Guid> CreateExperimentAsync(string name)
...
             using (var db = new LocalDbContext())
             {
-                var experiment = new Experiment(name);
-                await db.Experiments.AddAsync(experiment);
-                await db.SaveChangesAsync();
-                
-                return experiment.Id;
+	            var existingExperiment = db.Experiments.FirstOrDefault(x => x.ExperimentName == name);
+
+	            if (existingExperiment == null)
```
aslotte, line 21: Just a minor comment here about the indentation being one or two spaces too many

test/MLOps.NET.SQLite.IntegrationTests/SQLiteMetaDataStoreTests.cs
mbrdev: Should be fixed now

src/MLOps.NET.SQLite/Storage/SQLiteMetaDataStore.cs
mbrdev: Should be fixed now

aslotte:
Perfecto. Thank you very much sir!

PR #168, opened 2020-06-30
Added ability to log the data dimensions for the data used to train a model

### LINKED ISSUE #160: [Data] Add ability to capture data dimensions

**Additional context**

aslotte:
## Resolves
Fixes #160

## Description
Adds support to track column and schema for used data


aslotte:
The remaining work on this PR is to:
- Understand why GetRowCount comes back as null
- Implement IMetaDataStore methods in SQLite and Azure
- Add additional integration and unit tests

lqdev:
For RowCount, it looks like it might have something to do with whether the IDV is a file or not. When the IDV is created from an IEnumerable, you're able to get the RowCount. In part it kind of makes sense because you don't "Read" from the file until you've performed transformations. The IEnumerable, if it has already been instantiated already has the length/size information.

See this sample.

[data.txt](https://github.com/aslotte/MLOps.NET/files/4861570/data.txt)


```csharp
    class Program
    {
        static void Main(string[] args)
        {
            var dataEnumerable = new RandomData[]
            {
                new RandomData { HoursStudied=3, Score=80 },
                new RandomData { HoursStudied=0, Score=30 },
                new RandomData { HoursStudied=5, Score=97 }
            };

            var ctx = new MLContext();
            var idv = ctx.Data.LoadFromEnumerable(dataEnumerable);
            Console.WriteLine($"IEnumerable IDV: {idv.GetRowCount()}");

            var fileIdv = ctx.Data.LoadFromTextFile<RandomData>("data.txt",separatorChar:',',hasHeader:true);
            Console.WriteLine($"File IDV: {fileIdv.GetRowCount()}");
        }
    }

    class RandomData
    {
        [LoadColumn(0)]
        public float HoursStudied { get; set; }
        [LoadColumn(1)]
        public float Score {get;set;}
    }
```

IEnumerable IDV: 3
File IDV:

aslotte:
I think you're on to something @lqdev. I tried a couple of scenarios on Tuesday, e.g. calling `GetRowCount()` both after the data was loaded, after transformations and finally after the `Fit()` method was called. In neither case did it return a non-null value (this is for reading a text file).

To be honest, I'm quite surprised about this, especially as it didn't return something after calling the `Fit()` method, where I would have expected the estimator chain to have that information. May be an issue we need to add to the ML.NET repo, but for now I think it makes sense to row rowcount, as it is more confusing to explain to someone using the library why it's not there :)

lqdev:
I see what you mean and I would think so as well. Once it has transformed the data, I would expect the IDV to know at that point how many rows there are.

aslotte:
@AnoojNair  / @lqdev this is ready for a review.

PR #187, opened 2020-07-07
Added unit tests to LogMetricsAsync NaN check

aslotte:
## Resolves
NaN

## Description
Adds unit tests to NaN check for `LogMetricsAsync`

### DIFF
--- a/test/MLOps.NET.Tests/EvaluationCycleCatalogTests.cs
+++ b/test/MLOps.NET.Tests/EvaluationCycleCatalogTests.cs
@@ -1,0 +1,54 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using MLOps.NET.Catalogs;
+using MLOps.NET.Entities.Interfaces;
+using MLOps.NET.Storage;
+using MLOps.NET.Utilities;
+using Moq;
+using System;
+using System.Threading.Tasks;
+
+namespace MLOps.NET.Tests
+{
+    [TestCategory("UnitTests")]
+    [TestClass]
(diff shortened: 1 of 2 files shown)


aslotte:
@Brett-Parker feel free to review this PR :)

Brett-Parker:
@aslotte I was actually closer than I thought. Didn't think it was as easy as this;

`var metric = new { F1Score = double.NaN };` 

The act and assert I did the same. Just arrange I kept going round in circles. In future I will just have confidence and submit for comments :)

test/MLOps.NET.Tests/EvaluationCycleCatalogTests.cs
```diff
@@ -0,0 +1,54 @@
...
+using MLOps.NET.Utilities;
+using Moq;
+using System;
+using System.Threading.Tasks;
+
+namespace MLOps.NET.Tests
+{
+    [TestCategory("UnitTests")]
+    [TestClass]
+    public class EvaluationCycleCatalogTests
```
Brett-Parker, line 14: I think this has been discussed before but cannnot remember what the decision was. Should this be EvaluationCatalogTests? LifeCycle is a different class.

aslotte: Great catch, that was a typo from copy-pasting the old test class :)

aslotte:
Since we're using reflection to fetch the metric values, a dynamically created instance is sufficient, thus we can just use `var metric = new { F1Score = double.NaN };` 

You're always welcome to open PR's for discussion!

Issue #184, opened 2020-07-07
Add Cosmos DB integration tests using an emulator


aslotte:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Add integration tests similar to SQL Server and SQLite for our Cosmos DB metadata store: https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator#running-on-docker

aslotte:
Looking at the documentation, it appears that the current image can only be run in a Window container, which will not work for us (GitHub Actions require them to be able to run in a Linux container)

PR #188, opened 2020-07-08
Added initial CosmosDb Integration Test

aslotte:
## Resolves
Fixe #184 

## Description
Added initial implementation of CosmosDb integration tests. These currently only run locally against the Windows based Cosmos Db Emulator (https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator), but the CosmosDb team has in their roadmap a Linux version that will work with GitHub Actions in a couple of months.

Note that I've moved `EnsureCreated` to the `MLOpsBuilderExtension`. This is so we don't call this for each operation as it is redundant and has proven to be very slow for the CosmosDb implementation.

We will further need to consolidate these integration tests, as they will be very similar if not identical for the various storage providers. We still want to run them against different implementations though.


test/MLOps.NET.Azure.IntegrationTests/appsettings.json
```diff
@@ -1,5 +1,9 @@
 {
   "StorageAccount": {
     "ConnectionString": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;"
+  },
+  "Cosmos": {
+    "Endpoint": "https://localhost:8081",
+    "AccountKey": "C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw=="
```
aslotte, line 7: ℹ️ These are harmful, and the default connection values for the local emulator

aslotte:
This is ready for review!

AnoojNair:
@aslotte , looks like one of the unit test is failing because of the ensure created inside the builder.

aslotte:
Good catch, I totally missed that yesterday. After reviewing the unit tests, I actually decided to remove them. 
The reasoning being:
- They are covered the integration tests
- The dependencies in this scenario is not mockable, thus they are failing because they are doing real outbound calls -> not a unit test

aslotte:
Finally building...:)

PR #195, opened 2020-07-11
Added ability to get run by git commit hash

### LINKED ISSUE #194: Add ability to get run by Git Commit Hash
When setting up e.g. model tests after training has completed, we need the ability to fetch the latest run id that has been run. 
This will allow us to either download that specific model or other operations.

We want the ability to fetch a run by the git commit hash which we can associate with the model.

aslotte:
## Resolves
Fixes #194 

## Description
This method will be useful when we for example run model tests as we can fetch a run and the model by just using the git commit hash which we easily can get from the github actions workflow. We can then compare things such as model performance or run our own model tests.


aslotte:
This is ready for a review.

PR #198, opened 2020-07-12
Consolidated integration tests in to common place

### LINKED ISSUE #197: Consolidate integration tests
We currently have to add 3 exact same tests to test CosmosDb, SQLServer and SQLite leading to that we miss adding some tests and it takes longer to implement a new feature.

Keep all tests running for each storage provider, but consolidate them to only need to be written once.

aslotte:
## Resolves
Fixes #197 

## Description
This PR consolidates Cosmos DB, Sql Server and SQLite's integration tests into one place. The tests are going to be the same for all providers but we want them to run for each provider to make sure we don't miss any config issues or other special behavior. This change allows us to develop new features a lot faster as we now only need to add integration tests in one place and they will automatically be run for each provider.


aslotte:
@Brett-Parker  / @AnoojNair  - this is a change I've wanted to make for a while and thanks to all the previous refactoring we're finally here :) It will hopefully speed up development a lot. AFter this PR is merge, we only need to add functionality to one place in the MetaDataStore and it should work for all storage providers. Likewise we now also only need to add integration tests for that feature in one place and it will automatically be run for all providers. This will allow the library to have an increased quality as well as speed and scalability for new features.

PR #200, opened 2020-07-12
Refactored IMetaDataStore into smaller repositories

### LINKED ISSUE #199: Break IMetaDataStore apart into smaller Repositories
IMetaDataStore is rapidly growing. It's difficult to find which methods we've already implemented and where.

I would like the IMetaDataStore to be broken down to smaller repositories for entities, e.g.
- `IExperimentRepository`
- `IRunRepository`
- `IMetricRepository`
- `IDataRepository`

In addition, we may also want to break the `MetaDataStoreTests` into smaller test classes, I'm thinking one per catalog, e.g. `LifeCycleCatalogTests`, `EvaulationCatalogTests` and so forth to get a better view of what we're testing.

We should also think of if we really want methods to get metrics and hyperparameters or if it's enough just getting the run with all the associated items. It would be easier to maintain. (to be discussed)

aslotte:
## Resolves
Fixes #199 

## Description
Addresses the fact that the IMetaDataStore does too much. Breaking it down into smaller repositories instead so we can get a better overview

I also broke IMetaDataStoreTests into smaller tests per repo so we get a better view of what we actually tets.


aslotte:
Still need to break down the MetaDataStoreTests into a test file for each catalog (e.g. Evaluation, LifeCycle etc.)

aslotte:
@AnoojNair  / @Brett-Parker  / @dcostea this is another big refactor but should be the last one for now (I promise). The underlying structure is now much more streamlined and once we get used to it, it will be easier to add new features and also see what areas we have good testing on and which we don't.

Brett-Parker:
@aslotte don't apologize. Great work. I can't review right now but will take a look later if someone else hasn't already :)

PR #201, opened 2020-07-12
Added ability to automatically migrate the relational schema

### LINKED ISSUE #120: Add migrate ability for SQLite and SQL Server schema
When we update the schema, e.g. add a new property, our SQLite implemention is out of sync.

Add ability to automatically migrate the schema.

**Additional context**

aslotte:
## Resolves
Fixes #120 

## Description
`EnsureCreated` and `Migrate` are mutually exlusive, and `Migrate` will both create the database if it does not exist and migrate the schema if required. For now, we'll use automatic migration.

### DIFF
--- a/src/MLOps.NET/Storage/Database/MLOpsDbContext.cs
+++ b/src/MLOps.NET/Storage/Database/MLOpsDbContext.cs
@@ -38,7 +38,7 @@ namespace MLOps.NET.Storage.Database
         /// <returns></returns>
         public void EnsureCreated()
         {
-            Database.EnsureCreated();
+            Database.Migrate();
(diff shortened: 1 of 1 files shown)


aslotte:
This is ready for review.

aslotte:
Looking into the build failures. Doesn't seem to have been as easy as expected.

aslotte:
So I took a minute to read up on this, and from what I can tell, EF Core doesn't supported Automated Migrations as EF 6 did (which was an assumption I made). Looks like we would need to support migrations when changing database schemas, but I recommend doing that once we're out of the beta version :)

PR #217, opened 2020-07-15
Add ability to register model

### LINKED ISSUE #147: [Deployment] Add ability to register a model
For a given run, we would like to be able to register these model as something we would like to be able to deploy.
This issue is part of a larger deployment epic. We most likely will need a backing table with the following columns:

- Id
- RunId

With the following methods
- `RegisterModel(Guid runId)`
- 'RegisterModel(Guid runId, string filePath)'

**Additional context**
image

aslotte:
## Resolves
Fixes #147

### DIFF
--- a/src/MLOps.NET/MLOpsBuilder.cs
+++ b/src/MLOps.NET/MLOpsBuilder.cs
@@ -1,5 +1,6 @@
 using MLOps.NET.Storage;
 using MLOps.NET.Storage.Database;
+using MLOps.NET.Utilities;
 using System;
 
 namespace MLOps.NET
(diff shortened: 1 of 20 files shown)


aslotte:
This is ready for a review! Thanks all for helping out!

src/MLOps.NET/MLOpsBuilder.cs
```diff
@@ -34,7 +35,7 @@ public IMLOpsContext Build()
         public MLOpsBuilder UseMetaDataRepositories(IDbContextFactory contextFactory)
         {
             this.experimentRepository = new ExperimentRepository(contextFactory);
-            this.runRepository = new RunRepository(contextFactory);
+            this.runRepository = new RunRepository(contextFactory, new Clock());
```
mbrdev, line 38: Should this change be reverted since we didn't use it in the end?

aslotte: I think we can keep it. It will be useful once we start adding unit tests to the `RunRepository`

aslotte:
@mbrdev / @Brett-Parker - just realized Cosmos DB also most likely needed the common entity configuration that we added, so moved it to the common context. Ran our CosmosDB integration tests locally again and it all passed.

aslotte:
Thanks all!

PR #221, opened 2020-07-17
Create SQLServer Unit Test

### LINKED ISSUE #29: Add unit test project for Azure Storage Provider
We need some test automation to ensure that are properly setting values when creating experiments, runs and metrics

Brett-Parker:
## Resolves
Partially resolves #29 

## Description
Took the existing MLOpsBuilderExtensionTests.cs that was stored in the integration tests and added a new SQLServer unit test project to align with the other storage tests.


aslotte:
@Brett-Parker I think the MLOpsBuilderExtensionTests actually are integration tests, they for example create a connection to a SQL Server :) I think either we keep them in the integration test project, or we rewrite them to be real unit tests.

PR #227, opened 2020-07-22
[WIP] [Model] Added the ability to associate labels with a model 

### LINKED ISSUE #163: [Model] Add the ability to associate labels with a model
A need has been expressed where it would be useful to add labels to models, e.g. "LegalHasApproved" or "Experimental"

To support this, we'll need to add a new table called `Labels`.
We'll also need to add a new method `public void AddLabel(Guid runId, string name, string value)`

**Additional context**

Brett-Parker:
## Resolves
Fixes #163 

## Description
Added model label


Brett-Parker:
@aslotte Ready for review, along the right lines or not what you were thinking?

Brett-Parker:
I actually think this should be stored in the irun repo rather than its own.

aslotte:
What do you think @Brett-Parker, does it make the most sense to associate a label with an artifact or with an actual registered model? I'm thinking it may make sense to maybe make the association with the `RegisteredModel` (which I've yet to finalize the implementation for, sorry about that)

Brett-Parker:
The more I think about it. Labels are only relevant on the registered model imo. I think it's a nice to have elsewhere but probably not used as often. I committed this, went away and now think this isn't a good approach. Sorry if its taken up your time @aslotte

aslotte:
Not at all. We can keep this PR open if you like and re-purpose it once we got the RegisteredModel PR done (hopefully on Saturday). It's a great functionality to have later though.

aslotte:
@Brett-Parker I've changed the default branch to main, so given that this PR is on the old master branch I'm going to close it for. Feel free to apply the changes to the new main branch and open a new PR :)

PR #232, opened 2020-07-25
Adding RunRepository unit tests

aslotte:
## Resolves
N/A

## Description
I went ahead and added a unit tests to ensure that the RegisteredDate is actually properly set.

This is a fun PR as it showcases the use of:
- InMemoryDataBase for EF Core
- Named tuples in C# 8


aslotte:
@Brett-Parker  This is ready for  review. I just went ahead and added the unit test we spoke about during the stream.

PR #243, opened 2020-07-30
Ability to log data distribution

### LINKED ISSUE #162: [Data] Add ability to calculate and capture data distribution
For certain scenarios, e.g. classification models, the distribution of the data is important as it may greatly affect the model performance.

It would be cool if we could have the ability to calculate and track the data distribution of an IDataView (e.g. number of unique values in each column).

AnoojNair:
Fixes #162 

## Description
 @aslotte , i am trying to update the data entity with the data distribution and for some reason the data distribution which is nested 3 levels deep does not get inserted automagically. Data -> DataSchema -> DataColumn -> DataDistribution

Am i missing something here? Once i figure this out i will clean up rest of the PR.

            using var db = this.contextFactory.CreateDbContext();
            var data = db.Data.AsTracking().Include(d => d.DataSchema.DataColumns).First(d => d.RunId == runId);
            db.Entry(data).State = EntityState.Detached;  
            var dataColumn = data.DataSchema.DataColumns.FirstOrDefault(c => c.Name == columnName);
            var column = dataView.Schema.First(c => c.Name == columnName);
            List<DataDistribution> list = GetDataDistributionForColumn(dataView, columnName, column, dataColumn);
            dataColumn.Distribution.AddRange(list);
            db.Entry(data).State = EntityState.Modified;
            await db.SaveChangesAsync();


aslotte:
@AnoojNair could you clarify the exact issue you're seeing? Is it when you retrieve the `DataDistribution` that you've saved previously and it's coming back as null, or do you get an exception when attempting to save it?

I'm happy to pull the branch on your fork down later today to have a look if you could point me in the right direction. Concerning the method you're using to save the data, I think it can be simplified a lot. I don't think we need to modify the entity state or detach it from the context, but let me know what the issue you seeing is and I'll try to help.

AnoojNair:
@aslotte , The issue happens when i try to insert the DataDistributions for a column for the first time and since the DataDistributions is a nested property on Data , i was hoping it would be an Update on Data entity.

Did start with a simple implementation without the entity state but that didn't seem to work either. The code does not throw any exception but does not insert the Distribution into the DB either.

aslotte:
Hmm interesting. I can have a look at it tomorrow. I guess you've already removed your local database? Does it create a table for you? I'm writing this on my phone so can't check, but did you also add a DbSet to the context?

AnoojNair:
Yes, Added a new DbSet and it does create a Table with the right columns.

src/MLOps.NET/Storage/Repositories/DataRepository.cs
```diff
@@ -49,6 +52,51 @@ public async Task LogDataAsync(Guid runId, IDataView dataView)
             await db.SaveChangesAsync();
         }
 
+        ///<inheritdoc cref="IDataRepository"/>
+        public async Task LogDataDistribution(Guid runId, IDataView dataView, string columnName)
```
aslotte, line 56: Change this method to:
```
        public async Task LogDataDistribution(Guid runId, IDataView dataView, string columnName)
        {
            using var db = this.contextFactory.CreateDbContext();

            var data = db.Data
                .Include(d => d.DataSchema.DataColumns)
                .First(d => d.RunId == runId);

            var dataColumn = data.DataSchema
                .DataColumns
                .FirstOrDefault(c => c.Name == columnName);

            var column = dataView.Schema.First(c => c.Name == columnName);

            List<DataDistribution> list = GetDataDistributionForColumn(dataView, columnName, column, dataColumn);

            db.DataDistributions.AddRange(list);

            await db.SaveChangesAsync();
        }
```

test/MLOps.NET.IntegrationTests/DataCatalogTests.cs
```diff
@@ -72,6 +73,26 @@ public async Task LogDataAsync_ShouldGenerateDifferentHashWhenDataChanges()
...
+
+            //Act
+            await sut.Data.LogDataAsync(runId, data);
+            await sut.Data.LogDataDistribution(runId, data, "Sentiment");
+
+            //Assert
+            var savedData = sut.Data.GetData(runId);
+
+            savedData.DataSchema.DataColumns
+                .Single(x => x.Type == nameof(String) && x.Name == "Sentiment")
```
aslotte, line 91: This should be Boolean

src/MLOps.NET/Storage/Database/MLOpsDbContext.cs
```diff
@@ -15,6 +17,17 @@ public class MLOpsDbContext : DbContext, IMLOpsDbContext
...
             this.OnModelCreatingAction = configureEntityMap;
+
+            
+        }
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="entity"></param>
+        /// <returns></returns>
+        public override EntityEntry Entry(object entity)
```
aslotte, line 28: This can be removed but add

```
        DbSet<DataDistribution> DataDistributions { get; set; }
```

aslotte:
I forked your repo locally and got it to work. I think the initial issue was to add the list of data distributions to the `DbSet<DataDistribution> DataDistribution`. I couldn't find a property for it so added it and updated the DataRepository method and that seems to have made it work.

I added a couple of other comments :)

Is there a way to simplify the method to get data distribution? It would be awesome if we could avoid using reflection to fetch the distribution (if if gets too complex, we probably also want break that out to a separate service to be passed into the repo)

src/MLOps.NET/Storage/Repositories/DataRepository.cs
```diff
@@ -65,7 +113,7 @@ public Data GetData(Guid runId)
             using var db = this.contextFactory.CreateDbContext();
 
             var data = db.Data
-                .Include(x => x.DataSchema.DataColumns)
+                .Include(x => x.DataSchema.DataColumns).ThenInclude(x => x.Distribution)
```
aslotte, line 116: Update this method to

```
        public Data GetData(Guid runId)
        {
            using var db = this.contextFactory.CreateDbContext();

            var data = db.Data
                .Include(x => x.DataSchema.DataColumns)
                .FirstOrDefault(x => x.RunId == runId);

            data.DataSchema.DataColumns.ForEach(x => {
                x.Distribution = db.DataDistributions.Where(y => y.DataColumnId == x.DataColumnId).ToList();
            });

            return data;
        }
```

aslotte:
I hope it's helpful, great work!

AnoojNair:
Thanks @aslotte for looking into this. I will try this out...

AnoojNair:
@aslotte  , Would like to get some ideas on how to make the below much more simpler. Unfortunately the type for the 
GetColumn<> function on Idataview is only know at runtime.

```
var method = typeof(ColumnCursorExtensions).GetMethods().Single(m => m.Name == "GetColumn" &&
                      m.GetParameters()[1].ParameterType == typeof(string));

            var values = (dynamic)method
                        .MakeGenericMethod(column.Type.RawType)
                        .Invoke(dataView, new object[] { dataView, columnName });

```

aslotte:
Is there a way we can trigger the IDataView to read the data and then convert each column to an Enumerable to perform proper GorupBy?

Is there a way we can pass the IDataView to a DataFrame and use similar logic as built in there?

AnoojNair:
@aslotte , i have made the LogDataDistribution method generic so that the type parameter can be passed in from outside and that removes any need of reflection. What do you think?

aslotte:
@AnoojNair just wanted to comment that I'm planning to look at your question today or tomorrow :)

aslotte:
This is great @AnoojNair! I like that approach! 
Could we also change the distribution method to use 
`.Distinct()` to first get unique values and then loop over unique values and `Count` or alike to get unique values. I think that would be easier to read

AnoojNair:
@aslotte ,I will create a new PR so that the commit history is a bit more cleaner

PR #246, opened 2020-08-01
Adding support for URI deployment

### LINKED ISSUE #149: [Deployment] Add ability to deploy a model using an URI
This can certainly be up for debate, but I think we should add the ability to deploy a model via a URI so that a consuming application can listen to file changes. What we'll need to define for a deployment is which deployment environment we are going to deploy it to. If the deployment environment is of type URI, we should copy the model from the model-repository, over to a new container/folder (e.g. deployed-models) with the same Guid (Id) as the deployment environment has, meaning we should overwrite what's already there. 

We'll have to think about the security aspect in an upcoming issue, but for now the idea is that the model will be exposed with the path to the container, and the guid for the deployment environment.

We will at least need the following methods:
- `DeployModelAsync(DeploymentEnvironment environment, RegisteredModel model)`
- TBD

We will also need a backing table to keep track of deployed models, e.g. a table called DeployedModels with the following columns:
- Id
- DeploymentEnvironmentId
- RegisteredModelId
- DeploymentDate (DateTime)

This is part of a larger deployment epic.

image

aslotte:
## Resolves
Fixes #149 

## Description
- Adding support for URI deployment
- Adding support to get path to deployed model
- The model will be deployed in the same folder structure both for local file deploy, Azure and AWS
- The deployed model name will be the experiment name with a `.zip` ending and another deployment to the same deployment target will overwrite this file. This is so that a consuming application don't need to re-direct itself


aslotte:
An outstanding question from today's stream is how we want the URI to the model to be exposed.
Our alternatives are to have that defined on the DeploymentTarget, but that would ultimately tie a deployment target to an experiment and then cannot be re-used. We can also add a method otherwise for `GetDeploymentURI` which would figure out based on the registered model and deployment target what the URI or file path would be. That probably makes the most sense to be honest.

aslotte:
The next steps for this PR is:
- Add support to copy the registered model over to the new deployment container
- Implement that support for Azure Storage Container, AWS S3 and local file storage
- Add support to get the deployed file path/URI

aslotte:
Remaining work:
- Implement `DeployModel` for an S3 bucket
- Implement unit tests for the `DeploymentCatalog` `DeployModel` method
- Implement method to get the path to the deployed model based on `RegisteredModel` and `DeploymentTarget`
- Potentially a larger end-to-end test to create an experiment, create a run, register model, create deployment target etc and make sure that everything works as expected when we fetch our entities.

aslotte:
A remaining item that I forgot is that we cannot deploy the model with the RunId as it needs to support a versioned approach, meaning that if we deploy the next version, it should have the same name and overwrite the existing one

aslotte:
@Brett-Parker  / @AnoojNair  / Mitchell - this PR became a bit larger than expected, but I think it's all done and allows us to deploy a model via a URI now :) 

Let me know if you have any questions!

src/MLOps.NET.AWS/Storage/S3BucketModelRepository.cs
```diff
@@ -1,69 +1,114 @@
...
+        private readonly IAmazonS3 s3Client;
+        private readonly IModelPathGenerator modelPathGenerator;
+        private readonly string modelRepositoryBucket = "model-repository";
+        private readonly string deploymentRepositoryBucket = "deployment";
 
-        public S3BucketModelRepository(IAmazonS3 amazonS3Client, string bucketName)
-        {           
-            this.bucketName = bucketName;
-            this.amazonS3Client = amazonS3Client;
+        public S3BucketModelRepository(IAmazonS3 amazonS3Client, IModelPathGenerator modelPathGenerator)
```
Brett-Parker, line 19: `amazonS3Client` We changed this in to `s3Client` for consistency should we not change this too? Same in the rest of this file.

src/MLOps.NET.AWS/Storage/S3BucketModelRepository.cs
```diff
@@ -1,69 +1,114 @@
...
+
+            return GetDeploymentUri(deploymentTarget, registeredModel);
+        }
+
+        public string GetDeploymentUri(Deployment deployment)
+        {
+            return GetDeploymentUri(deployment.DeploymentTarget, deployment.RegisteredModel);
+        }
+
+        private string GetDeploymentUri(DeploymentTarget deploymentTarget, RegisteredModel registeredModel)
```
Brett-Parker, line 79: URI or URL?

aslotte: Uri :)

src/MLOps.NET.AWS/Storage/S3BucketModelRepository.cs
aslotte: Great catch, I'll push a change for that

test/MLOps.NET.AWS.IntegrationTests/S3ModelRepositoryTests.cs
```diff
@@ -43,5 +47,103 @@ public async Task UploadModelAsync_ShouldSucceed()
...
+            //Assert
+            var client = new HttpClient();
+            var response = await client.GetAsync(uri);
+
+            response.StatusCode.Should().Be(HttpStatusCode.OK);
+            response.Content.Headers.ContentLength.Should().BeGreaterThan(0);
+        }
+
+        [TestMethod]
+        public async Task DeployModelAsync_GivenADeployedModelAlreadyExist_ShouldOverwriteExistingModel()
```
Brett-Parker, line 111: `GivenADeployedModelAlreadyExist` - `GivenADeployedModelAlreadyExists`

test/MLOps.NET.Azure.IntegrationTests/StorageAccountModelRepositoryTests.cs
```diff
@@ -39,5 +53,103 @@ public async Task UploadModelAsync_ShouldSucceed()
...
+            //Assert
+            var client = new HttpClient();
+            var response = await client.GetAsync(uri);
+
+            response.StatusCode.Should().Be(HttpStatusCode.OK);
+            response.Content.Headers.ContentLength.Should().BeGreaterThan(0);
+        }
+
+        [TestMethod]
+        public async Task DeployModelAsync_GivenADeployedModelAlreadyExist_ShouldOverwriteExistingModel()
```
Brett-Parker, line 84: `GivenADeployedModelAlreadyExist` - `GivenADeployedModelAlreadyExists`

test/MLOps.NET.AWS.IntegrationTests/S3ModelRepositoryTests.cs
aslotte: 👍

test/MLOps.NET.Azure.IntegrationTests/StorageAccountModelRepositoryTests.cs
aslotte: 👍

test/MLOps.NET.Tests/Repositories/LocalFileModelRepositoryTests.cs
```diff
@@ -79,5 +84,45 @@ public async Task DownloadModel_ThrowsIfFileDoesNotExist()
...
+            // Act
+            var deployedPath = await sut.DeployModelAsync(deploymentTarget, registeredModel);
+
+            // Assert
+            deployedPath.Should().Be(expectedPath);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(InvalidOperationException), "The model to be deployed does not exist")]
+        public async Task DeployModel_NoSourceFileExist_ShouldThrowException()
```
Brett-Parker, line 114: `NoSourceFileExist` - `NoSourceFileExists`

aslotte:
Appreciate the thorough review @Brett-Parker! I've pushed corrections for all of your comments.

PR #268, opened 2020-08-10
Usability improvements and ensuring our object graph is resolved

### LINKED ISSUE #263: Usability improvements
* Return `RegisteredModel` from `RegisterModel`
* Return `DeploymentTarget` from `CreateDeploymentTarget`
* For `CreateDeploymentTarget` return existing if one by the same name already exists
* Rename `UseAWSS3Repository` to `UseAWSS3ModelRepository`

aslotte:
## Resolves
Fixes #263 

## Description
✔️ Return RegisteredModel from RegisterModel
✔️ Return DeploymentTarget from CreateDeploymentTarget
✔️ For CreateDeploymentTarget return existing if one by the same name already exists
✔️ Rename UseAWSS3Repository to UseAWSS3ModelRepository
✔️ Ensure that all related child entities are populated


aslotte:
I'm trying to create a simple solution to load the entire object graph for a retrieved entity so we don't return an object with null navigation properties.

### Challenges
- We cannot use `Include` as we use CosmosDb
- We cannot use Join as we use CosmosDb

### Attempt
I'm trying to create some kind of builder to resolve the object graph but hitting issues with stackoverflows as I've created nested dependencies. I think taking a step back and thinking of a better solution may be a good idea here

src/MLOps.NET/Storage/EntityResolvers/RunResolver.cs
```diff
@@ -0,0 +1,25 @@
+﻿using MLOps.NET.Entities.Impl;
+using MLOps.NET.Storage.Interfaces;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MLOps.NET.Storage.EntityResolvers
+{
+    internal sealed class RunResolver : IEntityResolver<Run>
+    {
+        public Run BuildEntity(IMLOpsDbContext db, Run entity)
```
aslotte, line 10: ℹ️ Dev note: I've added these resolvers so we can resuse them if we ever need to load these kind of entities, and they are at one place. This will make it easier to ensure we always correctly populate an entity

test/MLOps.NET.IntegrationTests/EndToEndTests.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace MLOps.NET.IntegrationTests
+{
+    public class EndToEndTests : RepositoryTests
+    {
+        [TestMethod]
+        public async Task DeployModel_GivenACompleteRun_ShouldDeployModel()
```
aslotte, line 13: ℹ️ Dev note: I've added this class in case we want to more cross-cutting tests, e.g. start to finish that doesn't drill into specifics but ensures that our flow is intact

aslotte:
@Brett-Parker  / @AnoojNair  would appreciate a review on this PR. 
If this looks okay, my next step and PR will be to change our models not to have a property for the parent object.
With that I mean removing `Run` from `Metric` and `Experiment` from `Run`. We would still keep the FK's to ensure data integrity but I believe this is a required step as we eager load everything and we may otherwise pull up a large object graph. It als simplifies our objects a lot.

aslotte:
Apologies, I didn't intend to push the last commit to this branch, but since it is done I'll do both changes in this PR. Moving it back to in progress.

aslotte:
We're getting closer today. There's some kind of SQLite issue that I will look at this afternoon.

aslotte:
@Brett-Parker  / @AnoojNair  this is now ready for your review. I realize this ended up being a really large PR and that's my bad. I normally want to keep it a lot smaller... :) 

In general, what this PR does is that it:
* Takes a good look at our DB schema and enforces FK where needed with proper on delete actions. 
* Removes the inverse navigation properties to the parent from a child so that when we eager load the model the payload will scale well
* It implements re-usable resolvers so that the entire object graph is loaded

I think there may be some more resolvers we can do, but I want to do that in a new PR so that this one doesn't grow even larger.
I also added an EndToEnd test that we can use to get a better picture of the full flow.

aslotte:
I'm going to merge this PR given that it contains migrations and changes that will cause merge conflicts for other contributions. I don't want to cause any conflicts on current or new PR's. Let me know if you have any comments or suggestions to this PR and I would be happy to open a new PR to address any of them :)

PR #270, opened 2020-08-11
Adding ability to log data distribution

### LINKED ISSUE #162: [Data] Add ability to calculate and capture data distribution

AnoojNair:
## Resolves
Fixes #162 

@aslotte , Like the idea of getting the distinct values and then counting . Makes the method much more readable.
Thank you :)


AnoojNair:
@aslotte , The SQL server test fails with 

Microsoft.Data.SqlClient.SqlException: Invalid object name 'DataDistribution'.

Do we have to force ensure created?

aslotte:
@AnoojNair  we have enabled migrations after the v1.0.0 release so you'll have to create a new EF code-first migration for both SQL Server and SQLite. I added some instructions on how to it here https://github.com/aslotte/MLOps.NET/wiki/Contributing-to-the-repository

aslotte:
Let me know if you have additional questions. Happy to update the docs so they make sense and are easy to follow :)

aslotte:
@AnoojNair  I forgot to mention in the instructions that there will be some generated code that requires XML comments (as we require it for all public methods). Either suppress them or add a dummy comment. I've updated the instructions.

AnoojNair:
@aslotte , Thank you.. the steps are very clear. added a dummy comment for now..

src/MLOps.NET.Azure/Storage/CosmosEntityConfigurator.cs
```diff
@@ -24,10 +24,12 @@ public static void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<RegisteredModel>().ToContainer(nameof(RegisteredModel));
             modelBuilder.Entity<DeploymentTarget>().ToContainer(nameof(DeploymentTarget));
             modelBuilder.Entity<Deployment>().ToContainer(nameof(Deployment));
+            modelBuilder.Entity<DataDistribution>().ToContainer(nameof(DataDistribution));
 
             modelBuilder.Entity<Data>().OwnsOne(x => x.DataSchema, dataSchema =>
             {
                 dataSchema.OwnsMany(x => x.DataColumns);
+                dataSchema.OwnsMany(x => x.DataColumns).OwnsMany(x => x.Distribution);
```
aslotte, line 32: I think an entity can't both own an entity e.g. DataColumns  -> owning Distribution and it also being it's own container (we will have to pick one). Given that it's logged separately from the Data, it may make sense that it is only in its own container

aslotte: I would be open to have it owned by a DataColumn too to be honest. I'll leave that for you to decide

src/MLOps.NET/Entities/Impl/DataColumn.cs
```diff
@@ -44,5 +45,10 @@ public DataColumn(Guid dataSchemaId)
         /// Type
         /// </summary>
         public string Type { get; set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public List<DataDistribution> Distribution { get; set; } = new List<DataDistribution>();
```
aslotte, line 52: If a `Distribution` is not owned by a `DataColumn` then this makes sense, otherwise we should remove it

src/MLOps.NET/Storage/Repositories/DataRepository.cs
```diff
@@ -69,7 +71,55 @@ public Data GetData(Guid runId)
...
+            var column = dataView.Schema.First(c => c.Name == columnName);
+
+            List<DataDistribution> list = GetDataDistributionForColumn<T>(dataView, column, dataColumn);
+
+            db.DataDistributions.AddRange(list);
+
+            await db.SaveChangesAsync();
+        }
+
+        private List<DataDistribution> GetDataDistributionForColumn<T>(IDataView dataView, DataViewSchema.Column column, Entities.Impl.DataColumn dataColumn) where T : struct
```
aslotte, line 103: ❤️ Much better!

aslotte: Is `T` a struct btw from `ML.NET`?

aslotte:
@AnoojNair  we may also want to add an `EntityMap` for `DataDistribution` to ensure that `Value` and `Count` are required. You may have to remove the previous migration and create a new one to add that

src/MLOps.NET.Azure/Storage/CosmosEntityConfigurator.cs
AnoojNair: I am happy with either, depends on how it will be used. If we have it separate then we would need a GetDistrubution mehod as well right?

aslotte: Exactly. I think keeping it as it's own entity makes sense for now. We should probably also make sure that any `GetData` method also loads these instances

AnoojNair:
@aslotte , i have separated DataDistribution from DataColumn and added a new GetDataDistribution method to the DataCatalog. Let me know if this looking ok.

src/MLOps.NET/Storage/Repositories/DataRepository.cs
```diff
@@ -71,5 +73,67 @@ public Data GetData(Guid runId)
...
+                .Include(d => d.DataSchema.DataColumns)
+                .First(d => d.RunId == runId);
+
+            var dataColumn = data.DataSchema
+                .DataColumns
+                .FirstOrDefault(c => c.Name == columnName);
+
+            var column = dataView.Schema.First(c => c.Name == columnName);
+
+            List<DataDistribution> list = GetDataDistributionForColumn<T>(dataView, column, dataColumn);
```
aslotte, line 92:
```suggestion
            var dataDistributions= GetDataDistributionForColumn<T>(dataView, column, dataColumn);
```

src/MLOps.NET/Storage/Repositories/DataRepository.cs
```diff
@@ -71,5 +73,67 @@ public Data GetData(Guid runId)
...
+
+            db.DataDistributions.AddRange(list);
```
aslotte, line 94:
```suggestion
            db.DataDistributions.AddRange(dataDistributions);
```

src/MLOps.NET/Storage/Repositories/DataRepository.cs
```diff
@@ -71,5 +73,67 @@ public Data GetData(Guid runId)
...
+
+            await db.SaveChangesAsync();
+        }
+
+        ///<inheritdoc cref="IDataRepository"/>
+        public List<DataDistribution> GetDataDistribution<T>(Guid runId, IDataView dataView, string columnName) where T:struct
```
aslotte, line 100: For this method I was not really thinking of exposing it to the `DataCatalog`. I was more thinking of making sure that the `DataDistributions` are loaded for the `GetData` method so when you get the data it will also be there :) 

Since `DataDistributions` are not an owned property, just remember that you'll need to get them by Id

src/MLOps.NET/Storage/Repositories/DataRepository.cs
```diff
@@ -71,5 +73,67 @@ public Data GetData(Guid runId)
 
             return data;
         }
+
+        ///<inheritdoc cref="IDataRepository"/>
+        public async Task LogDataDistribution<T>(Guid runId, IDataView dataView, string columnName) where T : struct
+        {
+            using var db = this.contextFactory.CreateDbContext();
+
+            var data = db.Data
```
aslotte, line 82: Let's add two checks here
1) The `Data` exist for a run (e.g. they have logged it already) and if not return an exception with a good message
2) The `DataColumn` exist for the data, and if not, return an exception again

src/MLOps.NET/Storage/Repositories/DataRepository.cs
```diff
@@ -71,5 +73,67 @@ public Data GetData(Guid runId)
...
+            var dataColumn = data.DataSchema
+                .DataColumns
+                .FirstOrDefault(c => c.Name == columnName);
+
+            var column = dataView.Schema.First(c => c.Name == columnName);
+
+            return GetDataDistributionForColumn<T>(dataView, column, dataColumn);
+        }
+
+        private List<DataDistribution> GetDataDistributionForColumn<T>(IDataView dataView, DataViewSchema.Column column, Entities.Impl.DataColumn dataColumn) where T : struct
```
aslotte, line 117: Could we move this to a separate service, e.g. `DataDistributionCalculator` which is passed into this repository. Like that you can add a lot more unit tests around this method to ensure that it does correctly calculate the distribution for various input data

aslotte:
@AnoojNair since I merged some changes for underlying data schema I took it on myself to try to help you out a bit by re-creating migrations and doing some changes according to my comments. I think there are still things that need to be fixed (I think I may have broken something). Happy to help you with it.

aslotte:
@AnoojNair all SQL Server and SQLite tests are passing but the Azure Cosmos DB tests are failing. I can try to help you look at that, I think may be something with the fact that DataColumn is owned but DataDistribution is not. It may be a bit tricky

aslotte:
@AnoojNair Cosmos tests fixed. I got tired of the issues we were having with owned entity types (which we didn't get much value from), so I changed all Data related entities to not owned :)

aslotte:
@AnoojNair I think everything should be resolved now. Feel free to review this PR. I didn't mean to take over the PR, but since kind of made it much more difficult for you merging some breaking changes beforehand, I thought I would give you a lending hand. I hope that was okay :) 

@Brett-Parker / @willvelida  feel free to review as well.

AnoojNair:
Thank you @aslotte for helping with the refactoring, i would have taken me quite a while , I will take a look at the PR .

AnoojNair:
@aslotte , Looks good to me. i don't think i can approve this as it is assigned to me.Happy for you approve and merge.

aslotte:
Awesome, thanks for reviewing it @AnoojNair and thank you for your contributions!

PR #283, opened 2020-08-15
Adding ability to add description to RegisteredModel

### LINKED ISSUE #220: Add description to saved model
There should be the ability to add a description to a trained model for deployment.

**Additional context**
 Its likely that the person training and deploying the model is not the same person consuming the model. The description can be a simple string field populated by the model trainer.

willvelida:
## Resolves

Fixes #220 by adding a string field for description to the ```RegisteredModel``` class.

## Description

In this PR, There is now a string property in the ```RegisteredModel``` class that allows users to add a basic description to the model. 

Unit Tests have also been updated.


willvelida:
@aslotte I've kept this simple and just added a string property for now. Let me know if you have any feedback :)

PR #285, opened 2020-08-15
Adding entity model schema diagram

aslotte:
## Description
Adding rough diagram for our entity model schema that can be used as a reference to understand how things are linked together.


aslotte:
@Brett-Parker  / @AnoojNair / @willvelida - this adds some simple schematics of how our entities are linked together.

PR #291, opened 2020-08-15
Added a RegisterModel method overload that does not require a description 

aslotte:
## Resolves
N/A

## Description
Added an overload to `RegisterModel` to allow a user to not needing to provide a description as the field is optional.


aslotte:
@willvelida added a simple method overload if a user don't want to provide a description :)

Issue #251, opened 2020-08-03
Add IsProduction flag on DeploymentTarget


aslotte:
**Is your feature request related to a problem? Please describe.**
Our deployment targets can have random names, based on what the users wants to name them. This is awesome but makes it difficult to e.g. write a method to fetch the evaluation metrics of the model in production.

**Describe the solution you'd like**
I propose we add a flag on deployment target for IsProduction that can be optionally set (default to false) when creating a deployment target.

willvelida:
@aslotte all good if I close this issue?

aslotte:
Please do!

PR #295, opened 2020-08-15
Adding IsProduction boolean flag to DeploymentTarget model.

willvelida:
## Resolves

Fixes Issue #251 

## Description

In this PR, the following changes have been made:

* ```IsProduction``` boolean property has been added to ```DeploymentTarget.cs``` file.
*  ```CreateDeploymentTargetAsync``` has been updated to allow an optional parameter to create a production deployment target in both ``DeploymentCatalog.cs``` and ```DeploymentRepository.cs```
* ```DeploymentTargetMap``` has been updated with the new ```IsProduction``` property along with migrations for both SQL and SQLite
* Updating Tests to check if default value for ```IsProduction``` is set to false and set to true when explicitly stated.
*


src/MLOps.NET/Storage/Repositories/DeploymentRepository.cs
```diff
@@ -47,6 +47,11 @@ public async Task<DeploymentTarget> CreateDeploymentTargetAsync(string deploymen
                 CreatedDate = clock.UtcNow
             };
 
+            if (isProduction)
```
aslotte, line 50: I don't think we need this if statement. We can just set the `IsProduction` flag directly :)

willvelida: Haha whoops 😊 Resolved it here: https://github.com/aslotte/MLOps.NET/pull/295/commits/b9d5bb7918d768a2ad379204dfc07a34de27071d

PR #329, opened 2020-09-05
Adding a MLOps.NET.CLI

### LINKED ISSUE #121: Add an MLOps.NET dotnet global tool

It would be interesting to have a dotnet global tool for some of the functionality in MLOps.NET so it can be automated via the CLI. That could include various deployment scenarios. 

A dotnet global tool is a command line nuget package. This is a good reference: https://www.dotnetcurry.com/dotnet/1495/dotnet-core-global-tools

**Additional context**

AnoojNair:
Fixes #121

@aslotte , Opened a new PR and closed the other 2.

Steps to install

Run dotnet pack where the MLOPS.Net.CLI.csproj file exists.
From the same location run
dotnet tool install --global --add-source ./nupkg mlops --version

Commands
mlops set-storage-provider --datasource "CosmosDb" --model-repository "S3"

Set endpoint and accountkey
mlops config-cosmos --endpoint "https://localhost:8081" --accountkey ""

Create experiment
mlops create-exp --name "test"

Create run
mlops create-run --exp-id ""

mlops ls-runs --exp-name "test"


aslotte:
Excellent work @AnoojNair! I made a couple of changes, I hope you don't mind :) 

1. Changes from inheritance to composition (e.g. removed the base class and instead injected the dependency)
I'm trying to avoid inheritance as much as possible and favor composition instead as it generally leads to more maintainable solutions once the solutions grows. Not a big difference, just a personal choice.

2. Removed Helper and called it just Cli. 
I'm trying to avoid `Helper` or `Service` as a name as much as possible. Naming is really hard, and I'm not sure if I name things well neither, but avoiding `Helper` and `Service` makes it less likely that classes grow outside what they were initially made to be if they are named more specifically :) 

3. Changed from package references to project references
I think this is what we want so that we can always stay up to date and add the latest functionality and methods to the CLI as well.

I haven't tried it locally yet, but I think this PR looks great and is good to merge! We can continue adding new commands and test it before releasing v1.2.0. Thank you for your contribution @AnoojNair and getting this `dotnet global tool` off the ground!

aslotte (approved):

AnoojNair:
@aslotte , Thanks for the refactoring , interfaces are a better option. With Regards to the project references, If in case we would like to publish the cli as a nuget package we would need to revert to package reference . What do you think?

aslotte:
I think that should be just fine. E.g., we use project references for all other internal dependencies and dotnet pack figures it out. For example MLOps.NET.SQLServer has a project reference to MLOps.NET and the dependency gets correctly listed. Happy to try it out locally later though

PR #345, opened 2020-09-12
[WIP] - Adding schema generator

### LINKED ISSUE #302: [Deployment] Add SchemaGenerator class to dynamically generate ModelInput and ModelOutput class definitions
When we will deploy a model in a container, we need to dynamically generate the class definitions of the model input and model output classes.

Create a `SchemaGenerator` class with a method `public string GenerateDefinition<T>(string className)`

The method should be able to generate and return a class definition as string including properties and custom attributes with values. The custom attributes we care about are `LoadColumn` and `ColumnName`.

**Update:** We probably want to call this method before deployment and store this in a data structure. E.g. I'm thinking `RegisterSchema<TInput, TOutput>()`

```
    public class ModelOutput
    {
        [ColumnName("PredictedLabel")]
        public bool Prediction { get; set; }

        [ColumnName("Score")]
        public float Score { get; set; }
    }
```

```
    public class ModelInput
    {
        [LoadColumn(0)]
        public bool Sentiment;

        [LoadColumn(1)]
        public string Review;
    }
```

**Pre-req** for #303

AnoojNair:
Fixes #302 

@aslotte , just wanted to have WIP PR to check if i am going in the right direction. Added a template folder for the various model output classes(build action: always copy) since the properties are same always. The Schema generator copies the model output from the template folder and returns a string version.

For Model input will have to think about a different strategy.


aslotte:
@AnoojNair it's certainly an alternative. What I'm afraid of is complex custom implementations or if a user just decides to use another name than e.g. `PredictedLabel` as the label name. What I was thinking is that we somehow decompile the runtime instance to source code, both for the model input and model output. After some googling, I think we may be able to use something like this: https://github.com/icsharpcode/ILSpy
https://stackoverflow.com/questions/29567489/can-roslyn-generate-source-code-from-an-object-instance

aslotte:
Let me know what you think, happy to continue bouncing some ideas as always. This is a tricky one :)

AnoojNair:
@aslotte ohh ok, decompiling the source code from th DLLs, have played around with ilspy a while back. Will try this out.. Thank you...

AnoojNair:
@aslotte , Updated the code to use ilspy and return the string version of the class from the application which calls this method.
Let me know if this will work?

src/MLOps.NET/Storage/Deployments/SchemaGenerator.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// 
+    /// </summary>
+    public sealed class SchemaGenerator
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="className"></param>
+        /// <returns></returns>
+        public string GenerateDefinition(string className)
```
aslotte, line 20: This is great @AnoojNair!

A couple of things:
- Can we pass in the type as a generic instead, e.g. `GenerateDefinition<T>(string className)` 
The reason I want to do that is className will be the name of the new class, but we need to fetch the name of class to base everything off from the type argument (they are going to be different names)

- Could we check the entire AppDomain instead of just the calling assembly? I think the calling assembly may just be `MLOps.NET` which is not going to contain this class, but if we look through all assemlies in the appdomain we should find it

test/MLOps.NET.Tests/Deployments/SchemaGeneratorTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        private SchemaGenerator sut;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            this.sut = new SchemaGenerator();
+        }
+
+        [TestMethod]
+        public void GenerateDefinition_ShouldGenerateModelOutput()
```
aslotte, line 27: Could we write a test here that checks that the outputed string matches **exactly** the class definition, including things like namespaces etc.? That would give us even more confidence that this will be doing what we need it to.

aslotte:
This is really exciting @AnoojNair! Added a couple of comments to think about.

src/MLOps.NET/Storage/Deployments/SchemaGenerator.cs
AnoojNair: @aslotte , Thanks for the feedback, I will make the changes.

AnoojNair:
@aslotte , I have made the changes.

test/MLOps.NET.Tests/Deployments/SchemaGeneratorTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            //Arrang
+
+            //Act
+            var modelOutput = sut.GenerateDefinition<ModelOutput>("BinaryClassificationModelOutput");
+
+            //Assert
+            modelOutput.Should().Contain("using Microsoft.ML.Data;");
+            modelOutput.Should().Contain("namespace MLOps.NET.Tests.Deployments");
+            modelOutput.Should().Contain("public class BinaryClassificationModelOutput");
+            modelOutput.Should().Contain("public bool Prediction");
```
aslotte, line 39: Nice! I'm happy to merge this for now and have a look at it more closely. One thing I think would be awesome is to split this class out until it's own file, and then assert each row (including e.g. `[ColumnName("PredictedLabel"]` etc.). This is a good start though if you just want me to merge it :)

src/MLOps.NET/Storage/Deployments/SchemaGenerator.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+namespace MLOps.NET.Storage.Deployments
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public sealed class SchemaGenerator
```
aslotte, line 14:
```suggestion
    internal sealed class SchemaGenerator
```

aslotte: Could you also real quick just add an interface to this class and move the summary comments to the interface? After that I happy to approve and merge the PR

aslotte: E.g. under an `Interfaces` folder

```
namespace MLOps.NET.Storage.Deployments
{
    /// <summary>
    /// Auto-generates a class based on run-time instance
    /// </summary>

    public interface ISchemaGenerator 
    {
        /// <summary>
        /// Searches the entire app domain to return the class definition as a string, using ilSpy
        /// </summary>
        /// <typeparam name="T"></typeparam>
        /// <param name="className"></param>
        /// <returns>Run-time instance's class definition</returns>
        string GenerateDefinition<T>(string className) where T: class;
    }
}
```

test/MLOps.NET.Tests/Deployments/SchemaGeneratorTests.cs
AnoojNair: done

aslotte (approved):
This is excellent work @AnoojNair and is going to be an integral part of auto-generating Docker images. Thank you for your contribution!

PR #378, opened 2020-10-01
Adding the ability to detect and save PackageDependencies for a Run

### LINKED ISSUE #316: Record package dependencies when creating a run
When deploying an ML.NET model in a container, we need to know what package dependencies were used during training. E.g. was a `Microsoft.ML.FastTree` package dependencies added? If these are not included in the deployment environment then we will most likely have a run-time issue.

Create a new data construct, `PackageDependency` with the following columns
- PackageDependencyId (Guid, PK)
- RunId (Guid, FK required)
- PackageName (string, required)
- PackageVersion (string, required)

Add a collection of package dependencies on a Run but no virtual navigation property to a Run from a PackageDependency (just the FK).

In the various `CreateRun` methods, discover all `Microsoft.ML` dependencies loaded into the app-domain and record them for a given run.

aslotte:
## Resolves
Fixes #316 

## Description
This PR adds the ability to automatically identify and log any `Microsoft.ML` dependencies during a run. This is important not only for auditability to know what version of ML.NET and what dependencies we used, but also as we can then install all of these dependencies automatically later when building out Docker container.

### DIFF
--- a/src/MLOps.NET.SQLServer/Migrations/20201001214039_AddPackageDependency.cs
+++ b/src/MLOps.NET.SQLServer/Migrations/20201001214039_AddPackageDependency.cs
@@ -1,0 +1,53 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace MLOps.NET.SQLServer.Migrations
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public partial class AddPackageDependency : Migration
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="migrationBuilder"></param>
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
+                name: "PackageDependency",
+                columns: table => new
+                {
+                    PackageDependencyId = table.Column<Guid>(nullable: false),
+                    RunId = table.Column<Guid>(nullable: false),
+                    Name = table.Column<string>(nullable: false),
+                    Version = table.Column<string>(nullable: false)
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("PK_PackageDependency", x => x.PackageDependencyId);
+                    table.ForeignKey(
+                        name: "FK_PackageDependency_Run_RunId",
+                        column: x => x.RunId,
+                        principalTable: "Run",
+                        principalColumn: "RunId",
+                        onDelete: ReferentialAction.Cascade);
+                });
+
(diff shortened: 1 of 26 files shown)


aslotte:
@AnoojNair  / @Brett-Parker / @willvelida  - this PR is ready for a review. Thanks!

src/MLOps.NET.SQLServer/Migrations/20201001214039_AddPackageDependency.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace MLOps.NET.SQLServer.Migrations
+{
+    /// <summary>
+    ///
```
Brett-Parker, line 7: Should these summaries be empty?

test/MLOps.NET.Tests/Catalogs/LifeCycleCatalogTests.cs
```diff
@@ -91,7 +98,29 @@ public async Task CreateRunAsync_WithoutGitCommitHash_ShouldProvideEmptyGitCommi
...
+        [TestMethod]
+        public async Task CreateRunAsync_WithDependencies_ShouldProvidePackageDependencies()
+        {
+            //Arrange
+            var dependencies = new List<PackageDependency>
+            {
+                new PackageDependency
+                {
+                    Name = "Microsoft.ML",
+                    Version = "1.5.2"
```
Brett-Parker, line 113: Does this mean we will need to update the version every update?

Brett-Parker (approved):
Good work. Few minor comments. Can ignore

src/MLOps.NET.SQLServer/Migrations/20201001214039_AddPackageDependency.cs
aslotte: This particular ones are just for the migrations, and as the solution as a whole enforces comments on public classes they are required but not really needed...so left them empty :)

test/MLOps.NET.Tests/Catalogs/LifeCycleCatalogTests.cs
aslotte: When we end up updating our ML.NET dependency yep. I couldn't find a better way to ensure that we actually read the proper dependencies than to use a real world use case for our tests. Hopefully it shouldn't be too much work, and be pretty obvious when it occurs.

aslotte:
Thank you for the review @Brett-Parker!

PR #380, opened 2020-10-02
[WIP] Issue360 register schema

### LINKED ISSUE #360: [Deployment] Add ability to RegisterSchema
We currently have a SchemaGenerator but no way to store it.

At the time at which someone would like to deploy a model to a container, they most likely don't have access to the input and output types. We need to store the types in a new table associated with the run.

I'm thinking a table called `ModelSchema` with columns:
- `ModelSchemaId`
- `RunId`
- `Name` (string, either ModelInput or ModelOutput)
- `Schema` (string)

I'm thinking a method called `RegisterSchema<TInput, TOutput>(Guid runId)`

AnoojNair:
## Resolves
Fixes #360


src/MLOps.NET/Storage/Database/MLOpsDbContext.cs
```diff
@@ -79,5 +79,10 @@ public virtual void EnsureCreated()
 
         ///<inheritdoc cref="IMLOpsDbContext"/>
         public DbSet<Deployment> Deployments { get; set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public DbSet<ModelSchema> Schemas { get; set; }
```
aslotte, line 86: I think we can remove this DbSet and instead add the ModelSchema a property to a Run instead (e.g. `public List<ModelSchema> ModelSchemas {get; set;}

src/MLOps.NET/Storage/EntityResolvers/ModelSchemaResolver.cs
```diff
@@ -0,0 +1,20 @@
+﻿using MLOps.NET.Entities.Impl;
+using MLOps.NET.Storage.Interfaces;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MLOps.NET.Storage.EntityResolvers
+{
+    internal sealed class ModelSchemaResolver : IEntityResolver<ModelSchema>
```
aslotte, line 8: I don't think we need this specific resolver, but we may want to add it to the `RunResolver`

src/MLOps.NET/Storage/Repositories/RunRepository.cs
```diff
@@ -169,5 +175,23 @@ public RegisteredModel GetLatestRegisteredModel(Guid experimentId)
...
             return this.registeredModelResolver.BuildEntity(db, latestRegisteredModel);
         }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="TInput"></typeparam>
+        /// <typeparam name="TOutput"></typeparam>
+        /// <param name="runId"></param>
+        public async Task RegisterSchema<TInput, TOutput>(Guid runId) where TInput : class where TOutput : class
```
aslotte, line 185: Let's do the schema resolution outside of the `RunRepository`. I want to try to keep the `RunRepository` focused on only managing the DB context. We can call the `SchemaGenerator` in e.g. the `LifeCycleCatalog` and pass the result in here.

src/MLOps.NET/Storage/Repositories/RunRepository.cs
```diff
@@ -169,5 +175,23 @@ public RegisteredModel GetLatestRegisteredModel(Guid experimentId)
...
+        {
+            using var db = this.contextFactory.CreateDbContext();
+            var modelInput = schemaGenerator.GenerateDefinition<TInput>("ModelInput");
+            var modelOutput = schemaGenerator.GenerateDefinition<TOutput>("ModelOutput");
+            var modelInputSchema = new ModelSchema(runId, typeof(TInput).Name, modelInput);
+            var modelOutputSchema = new ModelSchema(runId, typeof(TOutput).Name, modelOutput);
+            db.Schemas.Add(modelInputSchema);
```
aslotte, line 192: I think we should add a property to ModelSchema on a Run instance instead and in this method just get the run and update it by setting this property

src/MLOps.NET/Entities/Impl/ModelSchema.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+
+namespace MLOps.NET.Entities.Impl
+{
+    /// <summary>
+    /// Metric associated with a given training run
```
aslotte, line 6: Have a look at the comments in this class to make sure that they are correct

src/MLOps.NET/Storage/Repositories/RunRepository.cs
```diff
@@ -169,5 +175,23 @@ public RegisteredModel GetLatestRegisteredModel(Guid experimentId)
...
+        /// </summary>
+        /// <typeparam name="TInput"></typeparam>
+        /// <typeparam name="TOutput"></typeparam>
+        /// <param name="runId"></param>
+        public async Task RegisterSchema<TInput, TOutput>(Guid runId) where TInput : class where TOutput : class
+        {
+            using var db = this.contextFactory.CreateDbContext();
+            var modelInput = schemaGenerator.GenerateDefinition<TInput>("ModelInput");
+            var modelOutput = schemaGenerator.GenerateDefinition<TOutput>("ModelOutput");
+            var modelInputSchema = new ModelSchema(runId, typeof(TInput).Name, modelInput);
```
aslotte, line 190: Let's store the schema with the name ModelInput and ModelOutput rather than the class names

aslotte (commented):
Thanks for starting on this @AnoojNair! I know this PR is in progress, but wanted to leave a couple of comments early to nudge you in the right direction.

AnoojNair:
@aslotte Thanks for the early feedback, i have pushed the changes.

AnoojNair:
Will check the failed tests ..

aslotte:
Thanks @AnoojNair!
- I think you want to move the RegisterSchema method to the LifeCycleCatalog as it has to do with Runs
- Would you mind adding some integration tests to make sure the schema is converted and saved in the database?
- I think you'll also need to add migrations for SQLServer and SQLite

src/MLOps.NET/Storage/Repositories/RunRepository.cs
```diff
@@ -170,5 +173,21 @@ public RegisteredModel GetLatestRegisteredModel(Guid experimentId)
...
+
+       /// <summary>
+       /// 
+       /// </summary>
+       /// <typeparam name="TInput"></typeparam>
+       /// <typeparam name="TOutput"></typeparam>
+       /// <param name="runId"></param>
+       /// <param name="modelSchemas"></param>
+       /// <returns></returns>
+        public async Task RegisterSchema<TInput, TOutput>(Guid runId, List<ModelSchema> modelSchemas) where TInput : class where TOutput : class
```
aslotte, line 185: I don't think you need the TInput and TOutput here anymore :)

Issue #365, opened 2020-09-25
Update to .NET 5 and C# 9


aslotte:
**Is your feature request related to a problem? Please describe.**
We want to be using the latest technology.

**Describe the solution you'd like**
Update all projects to MLOps.NET to .NET 5

RonaldMariah:
@aslotte Happy to take this and create a PR.

aslotte:
Thanks @RonaldMariah - feel free to have a stab at it!

PR #384, opened 2020-10-03
Update to .NET 5 and C# 9

RonaldMariah:
….NET Core 3.1 packages to .NET 5 RC 1 versions.

## Resolves
Fixes issue #365 

## Description
Upgraded TargetFramework of .NET Core 3.1 projects to target .NET 5
Upgraded TargetFramework of .NET Standard class libraries to target .NET 5
Upgraded .NET Core 3.1 PackageReferences to .NET 5 RC 1 versions
Consolidated all Nuget packages with mismatching versions across all projects in MLOps.NET


RonaldMariah:
There's a failing integration test with the error:

Initialization method MLOps.NET.SQLite.IntegrationTests.KubernetesEndToEndTests.Initialize threw exception. System.InvalidOperationException: registryName cannot be empty.

@aslotte AFAIK this is in the settings file. I'm not sure if this gets injected from someplace that I'd need to change as well?

aslotte:
Thanks for your contribution so far @RonaldMariah. The specific test that is failing uses environment configuration values in the CI yml file (based on set secrets in this repo). They seem to still be set, but it's possible that something has changed in .NET 5 reading configuration values (or it may just be a Git fluke). I can look at it more in detail in the next few days.

PR #446, opened 2020-11-12
Target .NET 5.0 and C# 9

### LINKED ISSUE #365: Update to .NET 5 and C# 9

aslotte:
## Resolves
Fixes #365 

## Description
Target .NET 5.0 and C# 9

### DIFF
--- a/src/MLOps.NET/Docker/Interfaces/ICliExecutor.cs
+++ b/src/MLOps.NET/Docker/Interfaces/ICliExecutor.cs
@@ -14,15 +14,14 @@ namespace MLOps.NET.Docker.Interfaces
         /// <summary>
         /// Installs the dotnet new template package ML.NET.Templates
         /// </summary>
-        /// <param name="dockerSettings"></param>
         /// <returns></returns>
-        Task InstallTemplatePackage(DockerSettings dockerSettings);
+        Task InstallTemplatePackage();
 
         /// <summary>
         /// Adds package dependencies
         /// </summary>
         /// <returns></returns>
-        Task AddPackageDependencies(DockerSettings dockerSettings, List<PackageDependency> packageDependencies);
+        Task AddPackageDependencies(List<PackageDependency> packageDependencies);
 
         /// <summary>
         /// Uninstall the dotnet new template package for ML.NET.Templates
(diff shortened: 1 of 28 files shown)


aslotte:
@Brett-Parker this is ready to review

aslotte:
You'll need the .NET 5 SDK and the latest Visual Studio version to run it locally.

aslotte:
I'm investigating the failing integration test

aslotte:
The failing test `DeployModel_GivenACompleteRun_ShouldDeployModelToContainer` passes locally. It fail on the Linux agent due to the path to `image/ML.NET.Embedded.csproj` during the `dotnet add` command.

aslotte:
I'm very surprised, but it looks like the failing test that has been hindering this PR is due to the fact that running `dotnet new` returns before it is actually finished when running it through the CLI Wrapper package. This causes the following step to fail. I discovered this by adding some `Thread.Sleeps` but that's not a long term solution so I'm going to think about what to do next.

aslotte:
@Brett-Parker this is ready for review. I had to introduce a Thread.Sleep in one place which I hate, but not sure how to get around that `dotnet new --install` seems to return prior to the template package being installed.

aslotte:
I figured out a more robust version where we loop and check if the template is installed before proceeding :)

src/MLOps.NET/Docker/Interfaces/ICliExecutor.cs
```diff
@@ -14,15 +14,14 @@ public interface ICliExecutor
...
         /// </summary>
-        /// <param name="dockerSettings"></param>
         /// <returns></returns>
-        Task InstallTemplatePackage(DockerSettings dockerSettings);
+        Task InstallTemplatePackage();
 
         /// <summary>
         /// Adds package dependencies
         /// </summary>
         /// <returns></returns>
```
Brett-Parker, line 23: Should we add the parameters here? (Boy scouting)

```suggestion
        /// <returns></returns>
        /// <param name="packageDependencies"></param>
```

Brett-Parker (approved):
Looks good, I couldn't see anything to change. I picked up 1 tidy up but its just me boy scouting as I'm looking through. Great work, much needed job and gets us on the front foot moving forward!

aslotte:
Thanks @Brett-Parker, good catch! 👍 
I'm planning to prep a release tonight or tomorrow including this change and also Gregor's SQLite path change.

PR #499, opened 2020-12-20
Update ML.NET.Templates dependency to 0.4.0-beta

aslotte:
## Description
v0.4.0-beta runs on .NET 5, defaults to v1.5.4 of ML.NET and supports Swagger.


aslotte:
@Brett-Parker this is ready for your review. This makes sure that the Docker image we build runs on .NET 5 and the Web API supports Swagger which makes it easier to see what the expected model input and output will look like.

PR #716, opened 2021-08-12
Fix version conflicts caused by PR#713

sucrose0413:
I encountered a similar dependency hell issue when bump Microsoft.EntityFrameworkCore.Sqlite from 5.0.6 to 5.0.9. [(PR#713)](https://github.com/aslotte/MLOps.NET/pull/713).
Hope this fix can help you.

Best regards,
sucrose


aslotte:
Thanks @sucrose0413! I've been meaning to fix some of the dependency updates, but haven't had the time lately. The integration test that actually hits an Azure Kubernetes Instance may fail (as I've taken that service down temporarily, so I may need to fix that before this passes). Appreciate your contribution!

